PR #72, opened 2024-08-15
fix: fix minimum coder version warning

matifali:
It was incorrectly linked to 2.13.0.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -4,13 +4,13 @@ page_title: "coderd Provider"
 subcategory: ""
 description: |-
   ~> Warning
-  This provider is only compatible with Coder version 2.10.1 https://github.com/coder/coder/releases/tag/v2.13.0 and later.
+  This provider is only compatible with Coder version 2.13.0 https://github.com/coder/coder/releases/tag/v2.13.0 and later.
 ---
 
 # coderd Provider
 
 ~> **Warning**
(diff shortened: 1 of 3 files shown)


matifali:
@ethanndickson I don't know what Linter is unhappy about my PR.

ethanndickson:
@matifali Fixed. That was just `golangci-lint`, however the CI's using a newer version than is installed on our dogfood template.

ethanndickson:
The provider does work against Coder deployments going back to 2.10.1, even though those versions aren't technically supported.

Do we have any other projects that operate against a Coder deployment, that can be used against older, unsupported versions, absolutely fine? What Coder version do we say the software requires in those cases?

matifali:
Ok just change the link to 2.10.1 then for now. 
The PR mainly addresses the wrong link. Because clicking 2.10.1 takes you to 2.13.0. 
So just change the link for now and we can bump this version if/when we add new features that are only available in newer versions.

matifali:
Ok just change the link to 2.10.1 then for now. 
The PR mainly addresses the wrong link. Because clicking 2.10.1 takes you to 2.13.0. 
So just change the link for now and we can bump this version if/when we add new features that are only available in newer versions.

ethanndickson (approved):

PR #85, opened 2024-09-03
chore: validate fields as per coderd

ethanndickson:
Unfortunately, `coderd` validation errors don't return the reason why the validation failed, leading to confusing (almost useless) error messages when creating templates, groups or users with invalid names via the provider. 

The changes on `coderd/httpapi` to resolve this are non-trivial, as the [struct validation library](https://pkg.go.dev/github.com/go-playground/validator/v10) we use doesn't appear to support attaching an error reason to a field validator function.

Instead, we'll translate the string validation functions and regex from `coderd/httpapi/name.go`.

### DIFF
--- a/internal/provider/group_resource.go
+++ b/internal/provider/group_resource.go
@@ -77,13 +77,18 @@ func (r *GroupResource) Schema(ctx context.Context, req resource.SchemaRequest, 
 			"name": schema.StringAttribute{
 				MarkdownDescription: "The unique name of the group.",
 				Required:            true,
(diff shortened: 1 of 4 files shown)


ethanndickson:
* **#85** <a href="https://app.graphite.dev/github/pr/coder/terraform-provider-coderd/85?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

deansheather (approved):

PR #90, opened 2024-09-10
fix: correctly patch template acl

ethanndickson:
The provider currently does not remove users or groups from template ACL when applying a config.
This was unfortunately missed by tests, but caught by our use of the provider to manage our dogfood template.

Merely omitting a ACL item in a `coderd` API request is not enough to delete it, the role string must be set to an empty string. 
ACL items will now be deleted from the template when they are removed from the config and applied.

### DIFF
--- a/internal/provider/template_resource.go
+++ b/internal/provider/template_resource.go
@@ -531,7 +531,7 @@ func (r *TemplateResource) Create(ctx context.Context, req resource.CreateReques
 				if resp.Diagnostics.HasError() {
 					return
 				}
-				err = client.UpdateTemplateACL(ctx, templateResp.ID, convertACLToRequest(acl))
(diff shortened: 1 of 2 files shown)


ethanndickson:
* **#90** <a href="https://app.graphite.dev/github/pr/coder/terraform-provider-coderd/90?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

matifali:
Should we also provide/update the example to demonstrate this (if not done already)?

ethanndickson:
No, this PR just gives users the expected behavior that when they omit an ACL item from the list in their Terraform config, it will get deleted.

deansheather (approved):

PR #91, opened 2024-09-10
fix: downgrade coder v2.15.0 -> 2.14.2

ethanndickson:
I don't totally understand why CI tests passed (as goreleaser did not), but upgrading our codersdk `v2.15.0` introduced `coderdtest` as a dependency, when we want to avoid all the dependencies it requires, i.e. tailscale & gvisor, and the go mod replace directives for them.

 `codersdk` `v2.14.2` works fine against a `v2.15.0` deployment, so this is safe. When `v2.15.0` gets patched we'll upgrade once again.
 
 Related: https://github.com/coder/coder/pull/14633
 
 Treating as hotfix & merging.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.22.5
 
 require (
 	cdr.dev/slog v1.6.2-0.20240126064726-20367d4aede6
-	github.com/coder/coder/v2 v2.15.0
+	github.com/coder/coder/v2 v2.14.2
 	github.com/docker/docker v27.1.2+incompatible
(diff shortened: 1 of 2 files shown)


ethanndickson:
* **#91** <a href="https://app.graphite.dev/github/pr/coder/terraform-provider-coderd/91?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

ethanndickson:
My original intention was to downgrade so I could make a release with #90, but if we can patch the coder version this week I might just hold off.

PR #98, opened 2024-09-12
fix: create new template version when tfvars change

### LINKED ISSUE #97: Change on tf_vars should also trigger a new template version
Currently, a change in the tf_vars property does not trigger a recreation but a modification of the template.

As the tf_vars effectively change the content of the template, it should trigger a new version.

As an example, it would be possible to change the name of a parameter via tf_var.

A more drastic example to show what would be possible is to base the entire coder parameters on a tf_var:

template.tf
```hcl
resource "coderd_template" "example" {
  name        = "example"
  description = "example"

  versions = [
    {
      directory = "./template"
      active    = true
      tf_vars = [{
        name = "parameters"
        value = jsonencode([
          "cpu",
          "memory"
        ])
      }]
    }
  ]
}

```

in the template:
```hcl
variable "parameters" {
  type = string
}

locals {
  parameters = jsondecode(var.parameters)
}

data "coder_parameter" "parameters" {
  for_each = toset(local.parameters)
  name     = each.key
}
```

If I change the content of the parameters var via the coderd_template in this scenario, the whole template would change.

ethanndickson:
Closes #97.

### DIFF
--- a/internal/provider/template_resource.go
+++ b/internal/provider/template_resource.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"slices"
 	"strings"
 
 	"cdr.dev/slog"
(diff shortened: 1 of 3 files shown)


ethanndickson:
* **#98** <a href="https://app.graphite.dev/github/pr/coder/terraform-provider-coderd/98?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

internal/provider/template_resource.go
```diff
@@ -1231,8 +1232,9 @@ type LastVersionsByHash = map[string][]PreviousTemplateVersion
 var LastVersionsKey = "last_versions"
 
 type PreviousTemplateVersion struct {
-	ID   uuid.UUID `json:"id"`
-	Name string    `json:"name"`
+	ID     uuid.UUID `json:"id"`
+	Name   string    `json:"name"`
+	TFVars map[string]string
```
ethanndickson, line 1237: My understanding is that adding a new field to a struct is backwards compatible - if we read some private state from an older version of the provider the map will just be nil.

deansheather (approved):

PR #114, opened 2024-10-18
docs: replace example that requires TFC

ethanndickson:
The template example that used a Git commit hash in the name of each version referenced a Terraform variable that's only populated when using Terraform Cloud. There's not a great universal way to populate the variable within Terraform, so we'll instead mention two ways one could otherwise set it.


ethanndickson:
* **#114** <a href="https://app.graphite.dev/github/pr/coder/terraform-provider-coderd/114?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #123, opened 2024-10-31
fix: skip validating unknown version names

ethanndickson:
Same problem as seen in #115.

This fixes a problem where having multiple template version names manually set using Terraform variables would always return a duplicate name error, as Terraform would mark the attribute, although it is required and already written in the config, as Unknown before calling our custom validator. This unknown value is only ever passed to the validator, never `Create` or `Update`.

Knowing this, we see the same problem exists for all attributes we read in our custom validator. The only remaining one is `active`, where we check at least one version is marked as active. In this case, we also need to skip validating `active` if any of the booleans are unknown.

We previously didn't have any tests that explicitly used Terraform variables, so this PR adds one, in case the behaviour of these variable set attributes is changed in the future.

### DIFF
--- a/internal/provider/template_resource.go
+++ b/internal/provider/template_resource.go
@@ -896,9 +896,27 @@ func (a *activeVersionValidator) ValidateList(ctx context.Context, req validator
 		return
 	}
 
+	// Check all versions have unique names
+	uniqueNames := make(map[string]struct{})
+	for _, version := range data {
+		if version.Name.IsNull() || version.Name.IsUnknown() {
+			continue
+		}
+		if _, ok := uniqueNames[version.Name.ValueString()]; ok {
+			resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Template version names must be unique. `%s` appears twice.", version.Name.ValueString()))
+			return
+		}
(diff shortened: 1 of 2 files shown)


ethanndickson:
* **#123** <a href="https://app.graphite.dev/github/pr/coder/terraform-provider-coderd/123?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

internal/provider/template_resource.go
```diff
@@ -910,19 +928,6 @@ func (a *activeVersionValidator) ValidateList(ctx context.Context, req validator
...
 		resp.Diagnostics.AddError("Client Error", "At least one template version must be active.")
 	}
-
-	// Check all versions have unique names
-	uniqueNames := make(map[string]struct{})
-	for _, version := range data {
-		if version.Name.IsNull() {
-			continue
-		}
-		if _, ok := uniqueNames[version.Name.ValueString()]; ok {
```
ethanndickson, line 920: If two names were populated using Terraform variables, they'd be `Unknown` and `ValueString` would return the empty strin for both, which would then lead to a false duplicate in the map.

johnstcn (approved):

PR #126, opened 2024-11-01
chore: pin github actions to exact commits

aslilac:


.github/workflows/release.yml
```diff
@@ -1,37 +1,36 @@
...
+        uses: actions/checkout@v4
+      - name: Unshallow
+        run: git fetch --prune --unshallow
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: "1.22"
+      - name: Import GPG Key
+        id: import_gpg
+        uses: crazy-max/ghaction-import-gpg@cb9bde2e2525e640591a934b1fd28eef1dcaf5e5
```
ethanndickson, line 25: We should comment the hashes with `# vX.X.X` like we do in `coder/coder`. Dependabot notices and bumps the comment too.

.github/workflows/release.yml
```diff
@@ -1,37 +1,36 @@
...
-              env:
-                GPG_FINGERPRINT: ${{ steps.import_gpg.outputs.fingerprint }}
-                # GitHub sets this automatically
-                GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
+  goreleaser:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
ethanndickson, line 16: Might as well change them all?

ethanndickson (approved):
Thanks!

PR #130, opened 2024-11-05
fix: use `codersdk` functions for validating name attributes

aslilac:
rather than using regex that may drift over time, we should just use the functions provided by `codersdk` to validate name attributes


internal/provider/group_resource.go
```diff
@@ -77,17 +77,15 @@ func (r *GroupResource) Schema(ctx context.Context, req resource.SchemaRequest,
 				MarkdownDescription: "The unique name of the group.",
 				Required:            true,
 				Validators: []validator.String{
-					stringvalidator.LengthBetween(1, 36),
-					stringvalidator.RegexMatches(nameValidRegex, "Group names must be alpahnumeric with hyphens."),
+					codersdkvalidator.Name(),
```
ethanndickson, line 80: Group names need to use `codersdk.GroupNameValid` (which we recently bumped to like 255 characters)

internal/codersdkvalidator/validator_from_func.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+}
+
+var _ validator.Describer = functionValidator{}
+
+func (v functionValidator) Description(_ context.Context) string {
+	if v.err != nil {
+		return v.err.Error()
+	}
+	return "value must be a valid name"
```
ethanndickson, line 46: We should use the `defaultMessage` field here, right?

ethanndickson (approved):
Thanks for noticing this, this is much better! I actually moved the name validators from `coderd` into `codersdk` back in september and then forgot to use them here once the release came out :sob:

internal/codersdkvalidator/validator_from_func.go
aslilac: yes! good catch, I forgot the last step of this refactor lol

PR #49, opened 2024-07-29
feat: add coderd_organization resource

ethanndickson:
Terraform resource for creating and managing multiple organizations on a Coder deployment. On-hold until multi-orgs are complete.

Closes #39.

### Todo
- [ ] Organization custom roles
- [ ] Integration tests

PR #131, opened 2024-11-05
feat: add organization resource

aslilac:


ethanndickson:
I've got a bunch of boilerplate up on https://github.com/coder/terraform-provider-coderd/pull/49, feel free to steal and I'll close that PR

aslilac:
> I've got a bunch of boilerplate up on #49, feel free to steal and I'll close that PR. The test boilerplate and the member crud are probably the most useful

oh rad! thank you, I hadn't noticed that PR!

aslilac:
I filled out the meat of this PR using your other branch as a reference. I'll work on the tests tomorrow–again basing off of your branch, but I want to try to use the `statecheck` package that you mentioned on the issue.

aslilac:
I've removed `members` from the org resource because I have some different ideas of how I want that to work. If it doesn't work out we can always go back to this, but in the meantime this is a solid bit that I'm definitely happy with.

internal/provider/organization_resource.go
```diff
@@ -0,0 +1,229 @@
...
+	tflog.Trace(ctx, "successfully deleted organization")
+
+	// Read Terraform prior state data into the model
+	resp.Diagnostics.Append(req.State.Get(ctx, &data)...)
+}
+
+func (r *OrganizationResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	// Terraform will eventually `Read` in the rest of the fields after we have
+	// set the `id` attribute.
+	resource.ImportStatePassthroughID(ctx, path.Root("id"), req, resp)
```
ethanndickson, line 228: We should also support importing by organization name.

internal/provider/organization_resource_test.go
```diff
@@ -0,0 +1,132 @@
...
+					ConfigStateChecks: []statecheck.StateCheck{
+						statecheck.ExpectKnownValue("coderd_organization.test", tfjsonpath.New("name"), knownvalue.StringExact("example-org-new")),
+						statecheck.ExpectKnownValue("coderd_organization.test", tfjsonpath.New("display_name"), knownvalue.StringExact("Example Organization New")),
+					},
+				},
+			},
+		})
+	})
+
+	t.Run("CreateUnmanagedMembersOk", func(t *testing.T) {
```
ethanndickson, line 79: Should just delete this test, it's not in the schema and I don't forsee us adding it back. I think a separate resource for associating members is just straight up better.

internal/provider/organization_resource.go
```diff
@@ -0,0 +1,229 @@
...
+					codersdkvalidator.Name(),
+				},
+			},
+			"display_name": schema.StringAttribute{
+				MarkdownDescription: "Display name of the organization. Defaults to name.",
+				Computed:            true,
+				Optional:            true,
+				Validators: []validator.String{
+					codersdkvalidator.DisplayName(),
+				},
```
ethanndickson, line 72:
```suggestion
				Default: stringdefault.StaticString(""),
				},
```

I think I incorrectly had this (`display_name`) not using a `Default` field in my old org resource. This doesn't change the behaviour as `ValueString()` on an Unknown value (which it is if unse, and computed) produces empty string.

internal/provider/organization_resource.go
```diff
@@ -0,0 +1,229 @@
...
+			"id": schema.StringAttribute{
+				CustomType:          UUIDType,
+				Computed:            true,
+				MarkdownDescription: "Organization ID",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"name": schema.StringAttribute{
+				MarkdownDescription: "Username of the organization.",
```
ethanndickson, line 60:
```suggestion
				MarkdownDescription: "Name of the organization.",
```

Unless we've been using `username` for the organization unique name elsewhere?

internal/provider/organization_resource.go
aslilac: what if we actually _only_ supported importing by name?

ethanndickson: Is there any reason we wouldn't want to support both? Only supporting name would make it inconsistent with our other resources - is there any benefit?

ethanndickson: Whilst not quite the same, a lot of the resources in the GCP provider accept IDs in a bunch of different forms. 
https://registry.terraform.io/providers/hashicorp/google/6.11.1/docs/resources/gkeonprem_bare_metal_admin_cluster

Perhaps the lack of `x or y` imports is just because it's rare to have two totally unique identifiers for the same object on a SaaS.

internal/provider/organization_resource.go
```diff
@@ -0,0 +1,242 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+)
+
+// Ensure provider defined types fully satisfy framework interfaces.
+var _ resource.Resource = &OrganizationResource{}
+var _ resource.ResourceWithImportState = &OrganizationResource{}
+
+type OrganizationResource struct {
+	*CoderdProviderData
```
Emyrk, line 25: I'll just assume this data is thread safe, or is only accessed only by a single thread.

internal/provider/organization_resource.go
```diff
@@ -0,0 +1,242 @@
...
+
+func (r *OrganizationResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
+	// Read Terraform plan data into the model
+	var data OrganizationResourceModel
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	tflog.Trace(ctx, "creating organization")
```
Emyrk, line 156: Should we add log fields to log the org request name?

internal/provider/organization_resource.go
```diff
@@ -0,0 +1,242 @@
...
+		DisplayName: data.DisplayName.ValueString(),
+		Description: data.Description.ValueString(),
+		Icon:        data.Icon.ValueString(),
+	})
+	if err != nil {
+		resp.Diagnostics.AddError("Failed to create organization", err.Error())
+		return
+	}
+	tflog.Trace(ctx, "successfully created organization", map[string]any{
+		"id": org.ID,
```
Emyrk, line 168: org name as well?

internal/provider/organization_resource.go
```diff
@@ -0,0 +1,242 @@
...
+	var data OrganizationResourceModel
+	resp.Diagnostics.Append(req.State.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	orgID := data.ID.ValueUUID()
+
+	tflog.Trace(ctx, "deleting organization", map[string]any{
+		"id": orgID,
```
Emyrk, line 225: name here too?

Emyrk (approved):
LGTM :+1: 

I need to learn some of these terraform provider libs...

internal/provider/organization_resource_test.go
```diff
@@ -0,0 +1,114 @@
...
+						statecheck.ExpectKnownValue("coderd_organization.test", tfjsonpath.New("display_name"), knownvalue.StringExact("Example Organization")),
+						statecheck.ExpectKnownValue("coderd_organization.test", tfjsonpath.New("icon"), knownvalue.StringExact("/icon/coder.svg")),
+					},
+				},
+				// Import
+				{
+					Config:            cfg1.String(t),
+					ResourceName:      "coderd_organization.test",
+					ImportState:       true,
+					ImportStateVerify: true,
```
ethanndickson, line 63: `ImportState` uses the `id` attribute of the `ResourceName` by default. You can supply `ImportStateId` to override it to a specific value, i.e. an org name.

internal/provider/organization_resource.go
```diff
@@ -0,0 +1,242 @@
...
+	// Read Terraform prior state data into the model
+	var data OrganizationResourceModel
+	resp.Diagnostics.Append(req.State.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	var org codersdk.Organization
+	var err error
+	if data.ID.IsNull() {
```
ethanndickson, line 119: Your import test supplies a UUID into the name field, but `Organization` and `OrganizationByName` call the same endpoint, so I believe supporting imports by name will always allow you to import by ID? Setting the attributes at the bottom corrects the issue where the name is temporarily the UUID, so I don't think it's bug prone.

ethanndickson (approved):
👍

PR #132, opened 2024-11-05
chore: replace `PtrTo` with `ptr.Ref`

aslilac:
we already import coder/coder into this repo, and it's nice to have one name for things across code bases


ethanndickson (approved):
Not totally sure why we added `PtrTo` instead of importing `ptr.Ref` in the first place - I think maybe we were afraid of bringing in heavy imports cause it's in `coderd`? `util/ptr` doesn't import anything else so it's evidently fine 🤷

PR #134, opened 2024-11-08
chore: remove default org state

aslilac:
as an alternative to this, you should just do...

```terraform
data "coderd_organization" "default" {
  is_default = true
}
```

...and refer to `coderd_organization.default.id` when needed.

What we're currently doing is not _truly_ using the default organization, and hides details from the user that I think ultimately make the experience more dangerous and confusing.

### DIFF
--- a/internal/provider/template_resource.go
+++ b/internal/provider/template_resource.go
@@ -47,7 +47,7 @@ func NewTemplateResource() resource.Resource {
 
 // TemplateResource defines the resource implementation.
 type TemplateResource struct {
-	data *CoderdProviderData
+	*CoderdProviderData
 }
 
 // TemplateResourceModel describes the resource data model.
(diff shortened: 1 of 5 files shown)


internal/provider/template_resource.go
```diff
@@ -277,10 +277,10 @@ func (r *TemplateResource) Schema(ctx context.Context, req resource.SchemaReques
 				Default:             stringdefault.StaticString(""),
 			},
 			"organization_id": schema.StringAttribute{
-				MarkdownDescription: "The ID of the organization. Defaults to the provider's default organization",
-				CustomType:          UUIDType,
-				Optional:            true,
+				MarkdownDescription: "The ID of the organization that this template belongs to.",
 				Computed:            true,
+				Optional:            true,
```
ethanndickson, line 282: This should be `Required` now, and also on all the other resources.

ethanndickson (commented):
This is a breaking change, and should only get added when we're ready for the 1.0 release.

aslilac:
according to semver, every 0.0.x release is supposed to be considered "breaking".

we could even do 0.1.x, but I think this is something we should get feedback from long before 1.0, not afterwards.

aslilac:
- Allowing them to set a global default adds a lot of complexity to the provider
- It reduces our ability to validate the config during a plan (because `organization_id` would have to be optional on every resource that takes one)
- It's really not that hard for them to just add...

```terraform
data "coderd_organization" "default" {
  is_default = true
}

resource "..." "..." {
  organization_id = coderd_organization.default
}
```

It's a tiny bit of extra code for them, but avoids a lot of complexity, confusion, and magic for everyone.

ethanndickson:
I don't really agree that it adds a huge amount of complexity to the provider, user-facing or implementation wise, but I agree, it would be incredibly annoying for `terraform plan` to not pick up that you haven't set an org ID in either place. I've decided the implicit org selection (the first the user is a member of) is pretty bad, so yeah, let's just get rid of it for now, and then we can include it in the next release, where we're already bumping the minor version.

PR #141, opened 2024-11-15
feat: add `coderd_provisioner_key` resource

aslilac:
These are immutable once created, so really only `Create` and `Delete` need implementations. Everything gets stored in Terraform state (including the secret, since it cannot be reacquired any other way).

A change in any attribute must trigger a full replace of the resource.


internal/provider/provisioner_key_resource.go
```diff
@@ -0,0 +1,153 @@
...
+	// Read Terraform plan data into the model
+	var data ProvisionerKeyResourceModel
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	createKeyResult, err := r.Client.CreateProvisionerKey(ctx, data.OrganizationID.ValueUUID(), codersdk.CreateProvisionerKeyRequest{
+		Name: data.Name.ValueString(),
+		Tags: map[string]string{},
```
ethanndickson, line 109: Remember to set the tags

internal/provider/provisioner_key_resource_test.go
```diff
@@ -0,0 +1,96 @@
...
+	ctx := context.Background()
+	client := integration.StartCoder(ctx, t, "license_acc", true)
+	orgs, err := client.Organizations(ctx)
+	require.NoError(t, err)
+	firstOrg := orgs[0].ID
+
+	license := os.Getenv("CODER_ENTERPRISE_LICENSE")
+	if license == "" {
+		t.Skip("No license found for license resource tests, skipping")
+	}
```
ethanndickson, line 29: This check & skip already happens in `StartCoder` when is `useLicense` is true, so you can get rid of it. The reason we have it for the license resource tests is because we're adding the license ourselves via the new resource.

internal/provider/provisioner_key_resource.go
```diff
@@ -0,0 +1,153 @@
...
+
+	// Provisioner keys are immutable, no reading necessary.
+
+	// Save updated data into Terraform state
+	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
+}
+
+func (r *ProvisionerKeyResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
+	// Provisioner keys are immutable, updating is always invalid.
+	resp.Diagnostics.Append(diag.NewErrorDiagnostic("invalid update", "terraform is attempting to update a resource which must be replaced"))
```
ethanndickson, line 137:
```suggestion
	resp.Diagnostics.AddError("Invalid Update", "Terraform is attempting to update a resource which must be replaced")
```

`AddError` does the same thing here

internal/provider/provisioner_key_resource_test.go
```diff
@@ -0,0 +1,96 @@
...
+		IsUnitTest:               true,
+		PreCheck:                 func() { testAccPreCheck(t) },
+		ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: cfg1.String(t),
+			},
+			{
+				Config: cfg2.String(t),
+			},
```
ethanndickson, line 55: One test step modifying each of the attributes is probably good enough. You could also check the provisioner key attribute is populated with something that looks vaguely like a provisioner key.

ethanndickson (approved):
Resource is missing the markdown documentation, but CI is passing without `make gen` having been run 🤔 will fix

PR #147, opened 2024-11-25
feat: add `group_sync` and `role_sync` for `coderd_organization_resource`

aslilac:
Group and role sync settings can now be configured as such:

```tf
resource "coderd_organization" "example" {
	group_sync {
		field = "wibble"
		regex_filter = "wib{2,}le"
		auto_create_missing = true
		mapping = {
			wibble = ["6e57187f-6543-46ab-a62c-a10065dd4314"]
		}
	}

	role_sync {
		field = "wobble"
		mapping = {
			wobble = ["wobbly"]
		}
	}
}
```


internal/provider/organization_resource_test.go
```diff
@@ -71,6 +86,14 @@ func TestAccOrganizationResource(t *testing.T) {
 						statecheck.ExpectKnownValue("coderd_organization.test", tfjsonpath.New("display_name"), knownvalue.StringExact("Example Organization New")),
 					},
 				},
+				// Add group and role sync
+				{
+					Config: cfg3.String(t),
+					ConfigStateChecks: []statecheck.StateCheck{
+						statecheck.ExpectKnownValue("coderd_organization.test", tfjsonpath.New("group_sync.field"), knownvalue.StringExact("wibble")),
+						statecheck.ExpectKnownValue("coderd_organization.test", tfjsonpath.New("role_sync.field"), knownvalue.StringExact("wobble")),
```
ethanndickson, line 94:
```suggestion
						statecheck.ExpectKnownValue("coderd_organization.test", tfjsonpath.New("group_sync").AtMapKey("field"), knownvalue.StringExact("wibble")),
						statecheck.ExpectKnownValue("coderd_organization.test", tfjsonpath.New("role_sync").AtMapKey("field"), knownvalue.StringExact("wobble")),,
```

The name `AtMapKey` is a bit misleading, it's the `statecheck` way of simply appending to the path. Everything is either a slice step, or a map step.

internal/provider/organization_resource.go
```diff
@@ -83,6 +105,64 @@ func (r *OrganizationResource) Schema(ctx context.Context, req resource.SchemaRe
...
+						Optional: true,
+						MarkdownDescription: "Controls whether groups will be created if " +
+							"they are missing.",
+					},
+					"mapping": schema.MapAttribute{
+						ElementType:         types.ListType{ElemType: UUIDType},
+						Optional:            true,
+						MarkdownDescription: "A map from OIDC group name to Coder group ID.",
+						Validators: []validator.Map{
+							mapvalidator.ValueListsAre(listvalidator.ValueStringsAre(stringvalidator.Any())),
```
ethanndickson, line 139: The attribute is already `map(list(string))`, (UUIDType is also a string) so I don't think these validators are doing anything, though I assume they were added because `UUIDType` was broken.

internal/provider/organization_resource.go
```diff
@@ -83,6 +105,64 @@ func (r *OrganizationResource) Schema(ctx context.Context, req resource.SchemaRe
...
+						},
+					},
+					"mapping": schema.MapAttribute{
+						ElementType: types.ListType{ElemType: UUIDType},
+						Optional:    true,
+						MarkdownDescription: "A map from OIDC group name to Coder " +
+							"organization role.",
+						Validators: []validator.Map{
+							mapvalidator.ValueListsAre(listvalidator.ValueStringsAre(stringvalidator.Any())),
+						},
```
ethanndickson, line 161: Same here

internal/provider/organization_resource.go
```diff
@@ -133,6 +213,26 @@ func (r *OrganizationResource) Read(ctx context.Context, req resource.ReadReques
...
 	}
 
+	if !data.GroupSync.IsNull() {
+		_, err := r.Client.GroupIDPSyncSettings(ctx, data.ID.ValueUUID().String())
+		if err != nil {
+			resp.Diagnostics.AddError("Client Error", fmt.Sprintf("unable to get organization group sync settings, got error: %s", err))
+			return
+		}
+
+		// data.GroupSync = ???
```
ethanndickson, line 223: I don't recall if you can write to a config block (blocks can't be computed) like this. I suspect you can't.

If you can, I assume it would just be to error if the post-apply plan isn't empty.
For that you'd need to use `types.ObjectValueFrom` and provide it the type of the object using a `map[string]attr.Type`. I've got an example on the template resource & data source.

internal/provider/organization_resource.go
```diff
@@ -223,6 +339,19 @@ func (r *OrganizationResource) Update(ctx context.Context, req resource.UpdateRe
...
+		"orgID": orgID,
+	})
+	if !data.GroupSync.IsNull() {
+		r.patchGroupSync(ctx, orgID, data.GroupSync)
+	}
+	tflog.Trace(ctx, "updating role sync", map[string]any{
+		"orgID": orgID,
+	})
+	if !data.RoleSync.IsNull() {
+		r.patchRoleSync(ctx, orgID, data.RoleSync)
```
ethanndickson, line 352: Need to append diags, and return if error for both of these helpers.

internal/provider/organization_resource.go
aslilac: yeah, I was definitely just experimenting to see if I could jolt it into coercing the value out of that `unknown`

internal/provider/organization_resource.go
aslilac: I think I figured something out that works 😄

internal/provider/organization_resource.go
```diff
@@ -260,3 +473,78 @@ func (r *OrganizationResource) ImportState(ctx context.Context, req resource.Imp
...
+	var groupSyncData GroupSyncModel
+	diags.Append(groupSyncObject.As(ctx, &groupSyncData, basetypes.ObjectAsOptions{})...)
+	if diags.HasError() {
+		return diags
+	}
+
+	// Convert that into the type used to send the PATCH to the backend
+	var groupSync codersdk.GroupSyncSettings
+	groupSync.Field = groupSyncData.Field.ValueString()
+	groupSync.RegexFilter = regexp.MustCompile(groupSyncData.RegexFilter.ValueString())
```
ethanndickson, line 494: Keep this using `MustCompile`, but add a custom regex validator on each `regex_filter` attribute so it can be caught by `terraform validate`.

internal/provider/organization_resource.go
```diff
@@ -223,6 +417,25 @@ func (r *OrganizationResource) Update(ctx context.Context, req resource.UpdateRe
 		"icon":         org.Icon,
 	})
 
+	tflog.Trace(ctx, "updating group sync", map[string]any{
+		"orgID": orgID,
+	})
+	if !data.GroupSync.IsNull() {
+		resp.Diagnostics.Append(r.patchGroupSync(ctx, orgID, data.GroupSync)...)
```
ethanndickson, line 424: nit: do the `IsNull` checks in each `patch*` function

internal/provider/organization_resource.go
ethanndickson, line 1: I just realised we're missing enterprise/premium entitlement checks on this and the provisioner key resource. We have them on the workspace proxy and the template resource, but they run during `apply`, when they should really run during `plan`. Let's not worry about it for now, and at some point I'll see if I can get them working at plan-time using a plan modifier (validators don't have the provider config we need). 

If that doesn't work we can just copy the ones at apply-time to these two new resources.

internal/provider/organization_resource.go
```diff
@@ -83,6 +132,58 @@ func (r *OrganizationResource) Schema(ctx context.Context, req resource.SchemaRe
...
+					"regex_filter": schema.StringAttribute{
+						Optional: true,
+						MarkdownDescription: "A regular expression that will be used to " +
+							"filter the groups returned by the OIDC provider. Any group " +
+							"not matched will be ignored.",
+						Validators: []validator.String{
+							stringvalidator.LengthAtLeast(1),
+						},
+					},
+					"auto_create_missing": schema.BoolAttribute{
```
ethanndickson, line 156: This defaults to `false` on coder right? Should just have a `default` set in the schema, and then you can remove the null checks on it.

ethanndickson (approved):
few comments but otherwise nice, lgtm!

internal/provider/organization_resource.go
aslilac: would that prevent `group_sync` from being `null` tho?

internal/provider/organization_resource.go
aslilac: I like that this way makes it clear that we don't call those if nothing is set tho

internal/provider/organization_resource.go
ethanndickson: Nope, the default only gets computed if the parent is known, so:
```
    group_sync {}
```
would default the value, but omitting the block altogether would not. If you wanted to default the block to known, and the bool to false, then you'd also add a `Default` to the block itself.

PR #148, opened 2024-11-27
fix: mark invalid UUIDs as known

ethanndickson:
When I first implemented the UUID Terraform type, I followed the example from `terraform-provider-aws` for durations: [Source](https://sourcegraph.com/github.com/hashicorp/terraform-provider-aws/-/blob/internal/framework/types/duration.go). This example doesn't quite make sense, or at the least isn't applicable to our use-case, and since it was used as a base for our custom data type, resulted in UUIDs not being validated when written as strings in a configuration.

i.e. This function is called by the plugin SDK, when reading the config.
```
func (t durationType) ValueFromString(_ context.Context, in types.String) (basetypes.StringValuable, diag.Diagnostics) {
       [...]
	valueString := in.ValueString()
	if _, err := time.ParseDuration(valueString); err != nil {
		return DurationUnknown(), diags // Must not return validation errors
	}
       [...]
}
```
Meanwhile, this function is called during `validate`/`plan`/apply:
```
func (v Duration) ValidateAttribute(ctx context.Context, req xattr.ValidateAttributeRequest, resp *xattr.ValidateAttributeResponse) {
	if v.IsNull() || v.IsUnknown() {
		return
	}
}
```
This means if the duration couldn't be parsed, the value becomes unknown, and then is never validated, and so the provider just sees an Unknown value instead of an error diagnostic.

It's worth noting that you *cannot* remove the `IsUnknown` check from a validate function, as Terraform purposefully calls validate functions with unknown values when the attribute is set using a variable. The idea being to validate a config before and after any variables are populated. Therefore the only solution here is to return a *known*, *valid* value, if the UUID couldn't be parsed, as to let the validator handle it.

A similar data type is implemented in such a way that meets our use case for `arn` in `terraform-provider-aws`, which has been used as a guide for this fix: [Source](https://github.com/hashicorp/terraform-provider-aws/blob/0c9edd45ad5b079fe7186bf8ac957371a964811c/internal/framework/types/arn.go). In this provider, if an ARN can't be parsed, a valid & known value is returned from `ValueFromString`.

More importantly, I made the mistake of not testing the UUID type failing to parse in the context of a resource / data source. I only wrote internal unit tests for the type. This PR adds a test that ensures an error diagnostic is returned if an invalid UUID is provided to a resource/data-source UUID attribute, which tells us the problem is fixed.

### DIFF
--- a/internal/provider/user_data_source_test.go
+++ b/internal/provider/user_data_source_test.go
@@ -57,6 +57,7 @@ func TestAccUserDataSource(t *testing.T) {
 			Username: ptr.Ref(user.Username),
 		}
 		resource.Test(t, resource.TestCase{
+			IsUnitTest:               true,
 			PreCheck:                 func() { testAccPreCheck(t) },
 			ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
 			Steps: []resource.TestStep{
@@ -75,6 +76,7 @@ func TestAccUserDataSource(t *testing.T) {
 			ID:    ptr.Ref(user.ID.String()),
 		}
 		resource.Test(t, resource.TestCase{
+			IsUnitTest:               true,
 			PreCheck:                 func() { testAccPreCheck(t) },
 			ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
 			// User by ID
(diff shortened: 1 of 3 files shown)


ethanndickson:
* **#148** <a href="https://app.graphite.dev/github/pr/coder/terraform-provider-coderd/148?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/terraform-provider-coderd/148?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

PR #167, opened 2025-01-21
chore: allow only pushing inactive `coderd_template` versions

ethanndickson:


ethanndickson:
* **#167** <a href="https://app.graphite.dev/github/pr/coder/terraform-provider-coderd/167?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/terraform-provider-coderd/167?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

internal/provider/template_resource.go
```diff
@@ -1420,6 +1461,39 @@ func (planVersions Versions) reconcileVersionIDs(lv LastVersionsByHash, configVe
...
+	if !hasOneActiveVersion {
+		for i := range planVersions {
+			if !planVersions[i].ID.IsUnknown() {
+				prevs, ok := fullLv[planVersions[i].DirectoryHash.ValueString()]
+				if !ok {
+					continue
+				}
+				if versionDeactivated(prevs, &planVersions[i]) {
+					diag.AddError("Client Error", "Plan could not determine which version should be active.\n"+
+						"Either specify an active version or modify the contents of the previously active version before marking it as inactive.")
```
ethanndickson, line 1476: Instead of returning an error and failing the plan here, we could just push a new, identical, template version and not promote it to active. Given there's no other way to push an unmodified template version, and the whole point of this resource is to avoid pushing spurious diffs, I think it's better to just fail the plan.

internal/provider/template_resource.go
```diff
@@ -982,9 +992,37 @@ func (d *versionsPlanModifier) PlanModifyList(ctx context.Context, req planmodif
...
+				diags.AddError("Client Error", "Only one template version can be active at a time.")
+				return
+			}
+			active = true
+		}
+	}
+	if !active {
+		return false, diags
+	}
+	return true, diags
```
deansheather, line 1025:
```suggestion
	return active, diags
```

deansheather (approved):
Sounds good and the tests seem thorough. It's kinda hard to wrap your head around the one active version thing, but after I thought about it a bit your solution makes the most sense to me.

Issue #150, opened 2024-12-03
Support org sync


aslilac:
We've got group sync and role sync, but need to add org sync still

Emyrk:
Org sync is configured deployment wide, however for the terraform provider, it might be nice to configure it the `org` block.

```terraform
resource "coderd_organization" "this" {
  // ...
  idp_groups = ["developers"]
}
```

Unsure how ergonomic this is, just want to consider a different model than the ui is designed.

Kira-Pilot:
Will require us to implement https://github.com/coder/coder/issues/15939 first

PR #173, opened 2025-01-31
feat: add `coderd_organization_sync_settings` resource

aslilac:
Part of #150

Technically, this is complete by itself, as you can entirely configure organization sync this way, but this PR does not include the proposed "partial" API that we would still like to implement.

Example:

```terraform
resource "coderd_organization_sync_settings" "test" {
	field          = "wibble"
	assign_default = false

	mapping = {
		wobble = [
			"bf37c14a-cd36-4bf0-8ffb-e70d6ef8c518",
		]
	}
}
```


internal/provider/organization_sync_settings_resource.go
```diff
@@ -0,0 +1,265 @@
...
+			return diags
+		}
+	} else {
+		settings := codersdk.OrganizationSyncSettings{
+			Field:         field,
+			AssignDefault: assignDefault,
+			Mapping:       map[string][]uuid.UUID{},
+		}
+
+		// Terraform doesn't know how to turn one our `UUID` Terraform values into a
```
ethanndickson, line 215: nit:
```suggestion
		// Terraform doesn't know how to turn our `UUID` Terraform values into a
```

internal/provider/organization_sync_settings_resource.go
```diff
@@ -0,0 +1,265 @@
...
+	if err != nil {
+		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("unable to get organization sync settings, got error: %s", err))
+		return
+	}
+
+	// Store the latest values that we just fetched.
+	data.Field = types.StringValue(settings.Field)
+	data.AssignDefault = types.BoolValue(settings.AssignDefault)
+
+	if !data.Mapping.IsNull() {
```
ethanndickson, line 112: IIRC on import this will be unknown, not null, so `Mapping` *will* get populated. That sounds sensible to me?

internal/provider/organization_sync_settings_resource_test.go
```diff
@@ -0,0 +1,125 @@
...
+	cfg3.Mapping = map[string][]uuid.UUID{
+		"wibble": {uuid.MustParse("151b5a4e-391a-464d-a88c-ac50f1458d6f")},
+	}
+
+	t.Run("CreateUpdateReadOk", func(t *testing.T) {
+		resource.Test(t, resource.TestCase{
+			IsUnitTest:               true,
+			PreCheck:                 func() { testAccPreCheck(t) },
+			ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
+			Steps: []resource.TestStep{
```
ethanndickson, line 52: We should have an import test, it's a little hard to follow what the expected behaviour is

docs/resources/organization_sync_settings.md
ethanndickson, line 1: We should add an example to `examples/resources/coderd_organization_sync_settings/resource.tf` so it shows up on this doc page.

internal/provider/organization_sync_settings_resource.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+	tflog.Trace(ctx, "successfully deleted organization sync", map[string]any{})
+
+	// Read Terraform prior state data into the model
+	resp.Diagnostics.Append(req.State.Get(ctx, &data)...)
+}
+
+func (r *OrganizationSyncSettingsResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	// Any random string provided as the ID will work for importing.
+	resource.ImportStatePassthroughID(ctx, path.Root("field"), req, resp)
```
ethanndickson, line 264: I'm not sure what happens if the `field` retrieved from the server differs from the id used when importing. Does it complain about a state mismatch? Does the terraform state reflect the import id or the value from `Read`?

ethanndickson (approved):
I only have so much of an understanding of IdP sync settings, but this mostly looks lgtm! 
My only real comments are on how this resource handles getting imported.

aslilac:
I'm realizing that you don't actually _have_ to implement import, and I think skipping it is actually better than a janky import that's weird and has potential issues/adds complexity.

There can only ever be one of this type of resource (which does break the Terraform model a bit, but is also what our customers want) so importing isn't actually very helpful here.

PR #181, opened 2025-02-19
chore: fix golangci config

aslilac:
CI is broken right now because of some breaking change with golangci. Gotta update the config to fix things.


ethanndickson (approved):
Thanks for fixing!

PR #247, opened 2025-08-15
fix: add proper OIDC user role validation

angrycub:
When creating OIDC users, the provider was calling UpdateUserRoles even with empty roles due to the default schema value, causing the server error "User Role Field is set in the OIDC configuration".

OIDC users should get their roles exclusively from the OIDC provider's role mapping, not from explicit API calls. This fix:

- Errors if explicit roles are provided for OIDC users
- Skips role assignment entirely for OIDC users
- Provides clear error messaging about OIDC role behavior

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/internal/provider/user_resource.go
+++ b/internal/provider/user_resource.go
@@ -213,17 +213,28 @@ func (r *UserResource) Create(ctx context.Context, req resource.CreateRequest, r
 	resp.Diagnostics.Append(
 		data.Roles.ElementsAs(ctx, &roles, false)...,
 	)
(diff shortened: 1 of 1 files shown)


internal/provider/user_resource.go
ethanndickson, line 1: I think we need to update `Read` as well here, right? I don't have a Coder deployment w/ an IDP handy, but I assume if you gave the user managed by Terraform roles via OIDC, Terraform would complain about config drift on every subsequent apply.

PR #250, opened 2025-08-18
fix: support unmanaged roles on user resource

ethanndickson:
A customer ran into an issue when creating an OIDC user while role sync was enabled on the deployment:
```
Error: 'User Role Field' is set in the OIDC configuration. All role changes must come from the oidc identity provider. 
```
This is because we always call `UpdateUserRoles` on `Create` and `Update`.

OIDC User roles cannot be managed via the API if role sync is used, as the API always returns an error on any role update request. 

With this PR, `roles` can now be set to `null` in the config, whereby the Terraform provider will not attempt to read or update the user's roles under any circumstances. This prevents config drift when roles are set via Role Sync.


ethanndickson:
### Merge activity

* **Aug 19, 12:31 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/coder/terraform-provider-coderd/250).

PR #336, opened 2026-04-23
fix: Add `organization` option for workspace port sharing

angrycub:
Add missing `Organization` option for workspace sharing

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -146,6 +146,17 @@ func TestIntegration(t *testing.T) {
 			},
 		},
 		{
+			name: "template-port-share-test",
+			preF: func(t testing.TB, c *codersdk.Client) {},
+			assertF: func(t testing.TB, c *codersdk.Client) {
+				templates, err := c.Templates(ctx, codersdk.TemplateFilter{})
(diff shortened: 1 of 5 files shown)


johnstcn (commented):
Needs some corresponding tests in `internal/provider/template_resource_test.go`

Optionally, also a new integration test to validate the feature end-to-end.

angrycub:
Should I also add a test for `authenticated`, since it's not tested anywhere?

johnstcn:
> Should I also add a test for `authenticated`, since it's not tested anywhere?

Dealer's choice! I won't complain either way though 👍

michaelbrewer:
any updates?

angrycub:
Added two test cases:

- **`MaxPortShareLevelConstants`** (enterprise): steps through all four valid values (`owner`, `authenticated`, `organization`, `public`) on the same template resource, also exercising updates between values.
- **`InvalidMaxPortShareLevel`** (non-enterprise): confirms that an unrecognized value is rejected at plan time by the schema validator, before any API call is made.

ethanndickson (approved):

johnstcn (approved):

PR #344, opened 2026-05-03
feat: add archive_path attribute to coderd_template versions

### LINKED ISSUE #340: feat(coderd_template): Support also referencing an archive file.
Similar to the `coder templates push` command, the coderd_template resource should be able to optionally reference an archive file to push instead of just a directory. 

I have a use-case where I need to be able to include hidden directories as part of the template bundle, and to my understanding the way Coder handles it, it intentionally ignores them.  

In my use-case, I'm already filtering out the hidden files that I shouldn't be including via:
```
tar \
  --exclude='tmp.auto.tfvars.json' \
  --exclude='tf.plan' \
  --exclude='terraform.tfstate' \
  --exclude='.claude' \
  --exclude='.terraform' \
  --exclude='._*' \
  --exclude='.DS_Store' \
  --exclude='.git' -cvf - $(ls -A) | coder templates push -O coder memorycard --variables-file tmp.auto.tfvars.json -d -
```

But it would be nice if this was tracked and managed by the coderd_template instead.

jatcod3r:
Implements #340. Adds archive_path as an alternative to directory in the versions block of coderd_template, supporting .tar and .zip files.

Changes:
- Add archive_path (Optional) and archive_hash (Computed) attributes
- Validator ensures exactly one of directory/archive_path is set
- Plan modifier computes archive_hash from file contents (SHA-256)
- normalizeZip adds missing directory entries for archives produced by Terraform's archive_file data source
- uploadArchive sends the archive to Coder with correct content type
- Warn when switching between archive_path and directory about hidden file inclusion differences
- Document 100 MiB server upload limit in archive_path description

Tests:
- TestComputeArchiveHash (4 cases)
- TestArchiveContentType (5 cases)
- TestNormalizeZip (2 cases)
- TestReconcileVersionIDs extended with ArchiveHashMatching/Changed


ethanndickson:
@codex review

jatcod3r:
@codex review

jatcod3r:
@codex review

internal/provider/template_resource_test.go
ethanndickson, line 1: We should add a new subtest to `TestAccTemplateResource` that tests this feature works with a real coderd, if possible.

internal/provider/template_resource.go
```diff
@@ -456,12 +476,19 @@ func (r *TemplateResource) Schema(ctx context.Context, req resource.SchemaReques
...
-							Required:            true,
+							Optional:            true,
+							MarkdownDescription: "A path to the directory to create the template version from. Changes in the directory contents will trigger the creation of a new template version. Conflicts with `archive_path`.",
 						},
 						"directory_hash": schema.StringAttribute{
 							Computed: true,
 						},
+						"archive_path": schema.StringAttribute{
+							Optional:            true,
+							MarkdownDescription: "A path to a `.tar` or `.zip` archive file to upload as the template version source. Mutually exclusive with `directory`. Changes in the archive contents will trigger the creation of a new template version. The archive must not exceed 100 MiB (the Coder server upload limit).",
```
ethanndickson, line 470: We should probably validate the 100 MiB limit locally?

internal/provider/template_resource.go
```diff
@@ -1104,6 +1247,138 @@ func uploadDirectory(ctx context.Context, client *codersdk.Client, logger slog.L
...
+		return codersdk.ContentTypeTar, nil
+	case strings.HasSuffix(archivePath, ".zip"):
+		return codersdk.ContentTypeZip, nil
+	default:
+		return "", fmt.Errorf("unsupported archive format %q: must be .tar or .zip", filepath.Ext(archivePath))
+	}
+}
+
+// normalizeZip rewrites a zip archive to ensure that all intermediate directory
+// entries exist. Some tools (e.g. Terraform's archive_file data source) produce
```
ethanndickson: Given the edge case `archive_File` presents us, maybe we need an end-to-end test too that uses that? We have an integration test harness for the provider that should support running this sort of test. If it's a pain to do an acceptance test, like the one I mentioned in my other comment (I don't recall how easy it is), just an integration test would be fine.

internal/provider/template_resource.go
```diff
@@ -456,12 +476,19 @@ func (r *TemplateResource) Schema(ctx context.Context, req resource.SchemaReques
...
 						"directory": schema.StringAttribute{
-							MarkdownDescription: "A path to the directory to create the template version from. Changes in the directory contents will trigger the creation of a new template version.",
-							Required:            true,
+							Optional:            true,
+							MarkdownDescription: "A path to the directory to create the template version from. Changes in the directory contents will trigger the creation of a new template version. Conflicts with `archive_path`.",
 						},
 						"directory_hash": schema.StringAttribute{
 							Computed: true,
 						},
+						"archive_path": schema.StringAttribute{
```
ethanndickson, line 468: Neither me nor my agent can figure out why your agent used a new attribute here. Any ideas? Why couldn't we just use `directory_hash`? You could rename it but then you'd need a state migration, but it's not an attribute users would ever be concerned about so I think keeping it as `directory_hash` for both is fine?

jatcod3r: Did you mean to highlight `archive_hash`? Cause I agree, we could re-use `directory_hash`. Figured that I could detect changes based off the archive separately, but thinking about it now, we're really only worried about file content changes. Making that change now.

If you did mean archive_path, then I'm using that for targeting an individual archive file rather than a directory since that doesn't optionally cover standalone files. Unless u think we could modify it to encapsulate standalone archives too?

jatcod3r:
@codex review

jatcod3r:
@codex review

PR #355, opened 2026-06-05
feat: add default_org_member_roles to coderd_organization

Emyrk:
Refs [PLAT-217](https://linear.app/codercom/issue/PLAT-217/rfc-for-gateway-accounts), depends on coder/coder#25994.

Adds a `default_org_member_roles` set attribute to `coderd_organization` so callers can override the deployment-wide default member roles per organization.

- New attribute is Optional + Computed; omitting it leaves the deployment defaults in place.
- Create flow follows the `workspace_sharing` precedent: post-create `UpdateOrganization` PATCH when the user provides a value, since `CreateOrganization` doesn't accept the field.
- Update flow plumbs the value through the existing `UpdateOrganization` call as a `*[]string` so an unset attribute does not overwrite server state.
- Bumps `coder/coder` to a SHA on the gateway-accounts stack and reacts to coder/coder#24984, which migrated `UpdateTemplateMeta` fields from values to optional pointers.

<details><summary>Agent context</summary>

- `internal/provider/organization_resource.go`: new field on the resource model and schema, plus a `defaultOrgMemberRolesValueFromAPI` helper that maps a nil slice to an empty set so the attribute always has a known value.
- `internal/provider/organization_resource_test.go`: new acceptance step under `enableExperimentalSteps` that sets the field to `["organization-member", "organization-template-admin"]` and asserts state. The happy-path test now enables the `minimum-implicit-member` experiment alongside `workspace-sharing` since the server gates non-default values behind it.
- `internal/provider/template_resource.go` and `internal/provider/template_data_source_test.go`: react to coder/coder#24984's pointer migration on `UpdateTemplateMeta`. Mechanical conversions: `.ValueString()` becomes `.ValueStringPointer()` etc., and `bool` literals get wrapped with `ptr.Ref(...)`.
- `docs/resources/organization.md`: regenerated via `make gen`.

</details>

---

### DIFF
--- a/internal/provider/organization_resource.go
+++ b/internal/provider/organization_resource.go
@@ -45,6 +45,8 @@ type OrganizationResourceModel struct {
 	Icon             types.String `tfsdk:"icon"`
 	WorkspaceSharing types.String `tfsdk:"workspace_sharing"`
 
+	DefaultOrgMemberRoles types.List `tfsdk:"default_org_member_roles"`
+
 	OrgSyncIdpGroups types.Set    `tfsdk:"org_sync_idp_groups"`
 	GroupSync        types.Object `tfsdk:"group_sync"`
 	RoleSync         types.Object `tfsdk:"role_sync"`
@@ -147,6 +149,15 @@ This resource is only compatible with Coder version [2.16.0](https://github.com/
 				Validators: []validator.String{
 					stringvalidator.OneOf(workspaceSharingNone, workspaceSharingEveryone),
 				},
+			},
+
+			"default_org_member_roles": schema.ListAttribute{
+				ElementType: types.StringType,
+				MarkdownDescription: "Built-in organization role names that are unioned into every member's effective roles. " +
+					"Changes propagate to members on their next request. Setting any value other than the deployment defaults " +
+					"requires the `minimum-implicit-member` experiment to be enabled on the Coder Deployment.",
+				Optional: true,
+				Computed: true,
 			},
 
 			"org_sync_idp_groups": schema.SetAttribute{
@@ -352,6 +363,13 @@ func (r *OrganizationResource) Read(ctx context.Context, req resource.ReadReques
 	data.Icon = types.StringValue(org.Icon)
 	data.WorkspaceSharing = workspaceSharing
 
+	defaultOrgMemberRoles, diags := defaultOrgMemberRolesValueFromAPI(ctx, org.DefaultOrgMemberRoles)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	data.DefaultOrgMemberRoles = defaultOrgMemberRoles
+
 	// Save updated data into Terraform state
 	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
 }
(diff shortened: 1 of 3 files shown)


internal/provider/organization_resource.go
```diff
@@ -149,6 +151,15 @@ This resource is only compatible with Coder version [2.16.0](https://github.com/
...
 			},
 
+			"default_org_member_roles": schema.SetAttribute{
+				ElementType: types.StringType,
+				MarkdownDescription: "Built-in organization role names that are unioned into every member's effective roles. " +
+					"Changes propagate to members on their next request. Setting any value other than the deployment defaults " +
+					"requires the `minimum-implicit-member` experiment to be enabled on the Coder Deployment.",
+				Optional: true,
+				Computed: true,
+			},
```
Emyrk: Should this not be an array of strings?

internal/provider/organization_resource.go
```diff
@@ -456,6 +474,27 @@ func (r *OrganizationResource) Create(ctx context.Context, req resource.CreateRe
 		}
 	}
 
+	// Apply default_org_member_roles if the user specified them. CreateOrganization
+	// doesn't accept the field, so we follow up with UpdateOrganization.
```
Emyrk:
```suggestion
	// Apply default_org_member_roles if the user specified them.
```

internal/provider/organization_resource.go
Emyrk: Switched to a list. Verified the server preserves order (PostgreSQL `text[]`, no sort/dedupe in the handler), so no spurious diffs. Schema, model, helper, and test all updated; `make gen` now renders "List of String" in the docs.

internal/provider/organization_resource.go
Emyrk: Applied.

david-fraley (approved):

internal/provider/organization_resource.go
```diff
@@ -149,6 +151,15 @@ This resource is only compatible with Coder version [2.16.0](https://github.com/
 				},
 			},
 
+			"default_org_member_roles": schema.ListAttribute{
```
ethanndickson, line 154: Why is this a List and not a Set?

internal/provider/organization_resource.go
```diff
@@ -149,6 +151,15 @@ This resource is only compatible with Coder version [2.16.0](https://github.com/
...
+				ElementType: types.StringType,
+				MarkdownDescription: "Built-in organization role names that are unioned into every member's effective roles. " +
+					"Changes propagate to members on their next request. Setting any value other than the deployment defaults " +
+					"requires the `minimum-implicit-member` experiment to be enabled on the Coder Deployment.",
+				Optional: true,
+				Computed: true,
+			},
```
ethanndickson, line 161: This should have a validator

internal/provider/organization_resource.go
Emyrk: Set makes sense :+1: 

Will fix

internal/provider/organization_resource.go
Emyrk: @ethanndickson do you mean to validate the role names are valid?

At present we only support built in role names. So I can add this if you feel strongly.

If we add custom role support, then validating would be a lot more tricky.

ethanndickson: I actually don't think the session token from the provider block is available to us in the validator, so custom role validation is maybe impossible

Emyrk: Will covo move to the other PR

PR #357, opened 2026-06-05
feat(coderd_user): add service account support

### LINKED ISSUE #356: feat(coderd_user): Add service account support
### The gap

The Coder API and CLI support creating **service accounts**, but the `coderd` Terraform provider does not. `coderd_user` exposes `login_type` (`none | password | github | oidc`) but has no way to set the account as a service account, so it can't reach a capability the API already offers:

- API: `CreateUserRequestWithOrgs.ServiceAccount` (`json:"service_account"`)

- CLI: `coder users create --service-account`

- Server: the `users.is_service_account` column, the auto-assigned `organization-service-account` role, the `service_account:true` user filter, and the `is_service_account` field on user responses

The closest the provider can do today is `login_type = "none"`, which produces a login-disabled regular user — **not** a service account. The two are not equivalent:

- A `login_type = "none"` user (`is_service_account = false`) **counts toward the licensed user limit**, because active-user counting excludes service accounts but not login-disabled regular users. A real service account does not consume a seat.

- It receives the standard `organization-member` role instead of `organization-service-account`.

- It is not discoverable via the `service_account:true` filter or the `is_service_account` response field.

- It requires a non-empty email, whereas service accounts are the only users permitted an empty email.

As a result, there is no way to manage service accounts as code — they must be created out-of-band via the CLI/UI, which is the exact operational drift this provider is meant to close.

### Proposed change

- Add an `is_service_account` (bool, optional, defaults `false`) attribute to the `coderd_user` resource. The attribute name matches the API's `is_service_account` field (`json:"is_service_account"`), consistent with how `is_default` is named on the `coderd_organization` data source. On create it maps to the `service_account` field of `CreateUserRequestWithOrgs` (the create request uses the unprefixed name); this attribute-name-to-request-field mapping is the same approach already used for `suspended`.

- Mark it `ForceNew`: the service-account status is immutable server-side (set at creation and enforced by DB constraints), so changing it must recreate the user.

- Validate at plan time to mirror the server's rules: when `is_service_account = true`, default/require `login_type = none` and reject a non-empty `email` or `password`.

- Expose the same `is_service_account` attribute (read-only) on the `coderd_user` data source, populated from the response's `is_service_account` field — using one attribute name across both resource and data source, as the existing bool attributes do.

- Add an acceptance test and regenerate docs.

This closes the parity gap for service accounts without changing existing `coderd_user` behavior (the attribute is optional and defaults off).

We're happy to open the PR for this.

PushTheLimit:
Adds an `is_service_account` attribute to the `coderd_user` resource and data source, closing the Terraform/API parity gap for service accounts. Service accounts are admin-managed, login-less accounts that — unlike a `login_type = "none"` user — carry no email and do not consume a licensed user seat.

Changes:

- **Resource**: add `is_service_account` (optional, defaults `false`, `ForceNew` since it is immutable server-side). When `true`, the user is created via `CreateUserWithOrgs` with `ServiceAccount: true` (no email, `login_type` `none`); the existing `CreateUser` path is unchanged for regular users.
- **Plan-time validation** (`ValidateConfig`) mirrors the server rules: `email` is required unless `is_service_account` is `true`, in which case `email`, `password`, and a non-`none` `login_type` are rejected.
- **Data source**: expose `is_service_account` (read-only).
- `email` becomes `Optional + Computed` (was `Required`) so it can be omitted for service accounts; it remains effectively required for regular users via `ValidateConfig`.
- Add an acceptance test (`TestAccUserResourceServiceAccount`, gated on a licensed deployment via `UseLicense` since service accounts are Premium) and regenerate docs/examples.

The attribute name matches the API's `is_service_account` field, consistent with how `is_default` is named on the `coderd_organization` data source; on create it maps to the `service_account` field of `CreateUserRequestWithOrgs` (the create request uses the unprefixed name), the same attribute-name-to-request-field mapping already used for `suspended`.

Verified locally: `go build ./...`, `golangci-lint run` (clean), `make gen` (idempotent). Acceptance tests require a licensed deployment and run in CI.

Fixes #356


PushTheLimit:
**Implementation note:** In `ValidateConfig` I compare `login_type` against the `"none"` string literal rather than `codersdk.LoginTypeNone`. That constant is deprecated (staticcheck SA1019) in favor of service accounts, so referencing it fails the lint gate — and `"none"` is already how this resource expresses the value everywhere else (the schema default and the `OneOf` validator). I opted to respect the deprecation rather than add a `//nolint`, since suppressing a signal that points toward service accounts seemed at odds with what this change is adding. Happy to switch to `//nolint` + the typed constant if you'd prefer.

ethanndickson (approved):
Thanks for the PR!

PR #362, opened 2026-06-10
fix: handle unknown tf_vars and provisioner_tags at plan time

ericpaulsen:
## Problem

`coderd_template` fails during `ValidateResourceConfig` (and plan) with:

```
Value Conversion Error ... Received unknown value, however the target type
cannot handle unknown values. Path: [0].tf_vars,
Target Type: []provider.Variable, Suggested Type: basetypes.SetValue
```

whenever `tf_vars` or `provisioner_tags` are derived from module variables or sensitive values that Terraform marks as unknown at validate/plan time (for example `tf_vars = toset([for k, v in var.secrets : { name = k, value = v }])`).

The root cause is the data model: `TemplateVersion.TerraformVariables` and `TemplateVersion.ProvisionerTags` were plain Go slices (`[]Variable`). `ElementsAs` cannot decode an unknown set into a native slice, so both `versionsValidator.ValidateList` and `versionsPlanModifier.PlanModifyList` error out before the user ever reaches apply.

This was reported in #305 and is reproducible on `v0.0.16` and `v0.0.18`. #305 was auto-closed by #347, but #347 only fixed the plan-element rewriting path, not the `ValidateResourceConfig` path, so the bug is still present. **Please reopen #305** to track this.

## Fix

Change `TerraformVariables` and `ProvisionerTags` from `[]Variable` to `types.Set`, which can represent known, unknown, and null states. Native slices cannot represent Terraform's "unknown" sentinel, which is exactly the state these attributes take when their values flow from unknown variables.

Helpers (`variablesFromSet` / `variablesToSet` / `emptyVariableSet`) convert to and from `[]Variable` only when the set is known. Every consumer (`newVersion`, `setPrivateState`, `tfVariablesChanged`) unwraps the set defensively, treating null/unknown as "no known variables" (and, in `tfVariablesChanged`, treating an unknown set as changed so a new version is planned).

### Composition with #347

This change is complementary to #347, not an alternative:

- **#347** fixed plan-element rewriting in `PlanModifyList` (patch only `id`, `name`, and `directory_hash` on the original `attr.Value` elements instead of rebuilding the list with `ListValueFrom`, preserving cty sensitivity marks).
- **This PR** fixes the underlying data model so `ElementsAs` can decode unknown sets in the first place.

`PlanModifyList`'s element-patching logic is left untouched, so `tf_vars` / `provisioner_tags` keep their original `attr.Value`s (and their marks) on the plan.

## Credit

[...]

Issue #305, opened 2026-02-23
`coderd_template` "Attributes Set" input types issue


jatcod3r:
coder/coderd Version: v0.0.12

Running into the following error for both the `tf_vars` and `provisioner_tags` attributes
```
│ Error: Value Conversion Error
│ 
│   with module.ai-coder-com-coder.coderd_template.this,
│   on ../modules/template/main.tf line 77, in resource "coderd_template" "this":
│   77: resource "coderd_template" "this" {
│ 
│ An unexpected error was encountered trying to build a value. This is always an error in the provider. Please report the following to the provider developer:
│ 
│ Received unknown value, however the target type cannot handle unknown values. Use the corresponding `types` package type or a custom type that handles unknown values.
│ 
│ Path: [0].tf_vars
│ Target Type: []provider.Variable
│ Suggested Type: basetypes.SetValue
```

Full trace output:
```
2026-02-23T12:12:30.572-0800 [ERROR] provider.terraform-provider-coderd_v0.0.12: Response contains error diagnostic: tf_proto_version=6.9 tf_rpc=ValidateResourceConfig @module=sdk.proto diagnostic_attribute="ElementKeyInt(0).AttributeName(\"tf_vars\")" diagnostic_severity=ERROR diagnostic_summary="Value Conversion Error" tf_provider_addr=registry.terraform.io/coder/coderd tf_req_id=4c9eed24-af71-1c06-0acd-8d0342e91694 tf_resource_type=coderd_template @caller=github.com/hashicorp/terraform-plugin-go@v0.28.0/tfprotov6/internal/diag/diagnostics.go:58
  diagnostic_detail=
  | An unexpected error was encountered trying to build a value. This is always an error in the provider. Please report the following to the provider developer:
  | 
  | Received unknown value, however the target type cannot handle unknown values. Use the corresponding `types` package type or a custom type that handles unknown values.
  | 
  | Path: [0].tf_vars
  | Target Type: []provider.Variable
  | Suggested Type: basetypes.SetValue
   timestamp=2026-02-23T12:12:30.548-0800
[...]

PR #363, opened 2026-06-11
test(internal/provider): cover tf_vars and provisioner_tags passed via input variables

ericpaulsen:
Follow-up to #362, adding end-to-end regression coverage for #305 based on a customer-validated reproduction.

## Gap

`terraform plan` and `apply` run a validate walk first, in which **required input variables (no default) are evaluated as unknown**, even when a `-var-file` supplies concrete values. That walk calls `ValidateResourceConfig`, which is exactly where #305 blew up. No existing acceptance test could catch this class of bug: they all interpolate literal values into the config text, and the one variable-based test (`TestAccTemplateResourceSensitiveTFVarsDeferredReplan`) uses a variable with a `default`, which is known during the validate walk. The unit tests added in #362 cover the decode, but nothing exercised the real Terraform CLI path.

## Test

`TestAccTemplateResourceTFVarsFromVariable` mirrors the reported customer configuration: `tf_vars` and `provisioner_tags` supplied through required `list(map(string))` variables via `TestStep.ConfigVariables` (the harness equivalent of `terraform plan -var-file=...`), then a full plan, apply, and post-apply check. No license required.

Refs #305

<details>
<summary>Red/green validation and repro notes</summary>

- The customer config (`tf_vars = var.template_variables` with a required `list(map(string))` variable and `-var-file`) was reproduced verbatim against a provider built from `38cc6df` (pre-#362): `Value Conversion Error ... Path: [0].tf_vars, Target Type: []provider.Variable`, failing before the provider was even configured.
- The same config against merged `main` plans, applies against a local Coder v2.34.1 container, lands the variables on the created template version (verified via API), and re-plans empty.
- This test run against `38cc6df`: **fails** with `Value Conversion Error ... Path: [0].provisioner_tags, Target Type: []provider.Variable` (same decode; whichever field errors first).
- This test run against `main` with `TF_ACC=1`: **passes** in ~6.5s.
- The variables in the test intentionally have no defaults; adding a default would make them known during the validate walk and silently skip the regression.

</details>

> Disclosure: this pull request was generated by Coder Agents on behalf of @ericpaulsen.

### DIFF
--- a/internal/provider/template_resource_test.go
+++ b/internal/provider/template_resource_test.go
@@ -1194,7 +1194,9 @@ resource "coderd_template" "test" {
 // fields this failed with "Value Conversion Error ... Path: [0].tf_vars"
 // before the provider was even configured. The variables must not have
 // defaults; a default would make them known during the validate walk and skip
-// the regression entirely.
+// the regression entirely. The two variables deliberately use different type
+// constraints, list(object) and list(map(string)), to cover both input shapes
+// from the original reports.
 func TestAccTemplateResourceTFVarsFromVariable(t *testing.T) {
 	t.Parallel()
 	if os.Getenv("TF_ACC") == "" {
@@ -1208,7 +1210,10 @@ provider coderd {
 }
 
 variable "template_variables" {
-	type = list(map(string))
+	type = list(object({
+		name = string,
+		value = any
+	}))
 }
 
 variable "provisioner_tags" {
--- a/internal/provider/template_resource_test.go
+++ b/internal/provider/template_resource_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-testing/config"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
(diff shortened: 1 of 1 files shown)


internal/provider/template_resource_test.go
```diff
@@ -1185,6 +1186,86 @@ resource "coderd_template" "test" {
...
+	}
+
+	cfg := `
+provider coderd {
+	url   = %q
+	token = %q
+}
+
+variable "template_variables" {
+	type = list(map(string))
```
RobertClarke64: This is more specific, though either should work

```suggestion
	type = list(object({
		name = string,
		value = any
	}))
```

ericpaulsen: Applied in 84b2d87, with co-author credit. Verified both type constraints behave identically for this regression: built the provider at pre-fix `38cc6df` and ran this test with `list(object({name = string, value = any}))`, and it fails with the same `Value Conversion Error ... Path: [0].tf_vars`; on current `main` it passes.

Also confirmed your earlier question about mixed value types end to end: with `value = any` and a var-file containing `value = false` alongside `value = "coder"`, Terraform unifies the list to strings, and apply lands `use_kubeconfig = "false"` on the created template version. Both `list(map(string))` and the more specific `list(object)` form work. The `provisioner_tags` variable in the test deliberately keeps `list(map(string))` so both input shapes stay covered.

> Disclosure: this comment was generated by Coder Agents on behalf of @ericpaulsen.

ethanndickson (approved):

PR #368, opened 2026-06-23
feat: add experimental coderd_ai_provider resource

ethanndickson:
> **Stack** — builds on #370 (external Postgres sidecar for acceptance tests); review/merge that first.

## Summary

Adds `coderd_ai_provider` for managing Coder AI Gateway providers from Terraform — OpenAI-style API-key providers and AWS Bedrock providers — with full CRUD + import, generated docs/examples, and acceptance/unit tests. The resource is marked experimental via a warning callout in its docs.

## Approach

- **Write-only secrets.** Plaintext API keys and AWS credentials use Terraform 1.11+ write-only arguments (`*_wo`), so secrets reach Coder but never land in state. Each secret is paired with a `*_version` field; bumping the version rotates the secret. A null version means "unmanaged/preserve" rather than "clear", so removing it never silently wipes server-side secrets.
- **Split validation.** Schema validators handle the simple shape (required-together, non-empty); resource-level `ValidateConfig` handles type-dependent combinations (e.g. `api_key_wo` is rejected for `bedrock`/`copilot` or any `settings.bedrock` config, mirroring the server) and Bedrock's region-or-credentials requirement. Validation defers while values are unknown during the validate/plan walk.
- **Bedrock region.** When `region` is omitted it derives from a canonical Bedrock `base_url` (`bedrock-runtime.<region>.amazonaws.com`), so it re-derives correctly when `base_url` changes regions.

## Schema

```hcl
resource "coderd_ai_provider" "openai" {
  type               = "openai"
  name               = "openai"
  base_url           = "https://api.openai.com"
  api_key_wo         = var.openai_api_key
  api_key_wo_version = 1
}

resource "coderd_ai_provider" "bedrock" {
  type     = "bedrock"
  name     = "aws-bedrock"
  base_url = "https://bedrock-runtime.us-east-1.amazonaws.com"

  settings = {
    bedrock = {
      model                  = "anthropic.claude-3-5-sonnet-20241022-v2:0"
      access_key_wo          = var.bedrock_access_key
      access_key_secret_wo   = var.bedrock_access_key_secret
      credentials_wo_version = 1
    }
  }
}
```

Computed: `id`, `display_name`, `enabled`, `api_key_masked`, `created_at`, `updated_at`, and `settings.bedrock.region` (derived from `base_url`).

## Notes

Requires Terraform 1.11+ when configured (write-only arguments); acceptance cases skip earlier versions.

An integration test combining this with the model resource is planned.

Relates to CODAGT-607


ethanndickson:
@codex review

ethanndickson:
@codex review

internal/provider/ai_provider_resource.go
```diff
@@ -0,0 +1,595 @@
...
+		if accessSet {
+			accessKey := bedrock.AccessKeyWO.ValueString()
+			sdkSettings.AccessKey = &accessKey
+		}
+		if secretSet {
+			accessKeySecret := bedrock.AccessKeySecretWO.ValueString()
+			sdkSettings.AccessKeySecret = &accessKeySecret
+		}
+		if !sdkSettings.IsConfigured() {
+			resp.Diagnostics.AddAttributeError(path.Root("settings").AtName("bedrock"), "Missing Bedrock Settings", "`type = \"bedrock\"` requires Bedrock settings sufficient for the Coder API: set `region` or write-only AWS credentials.")
```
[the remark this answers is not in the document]
ethanndickson: Verified this against `coder/coder` at current `main` — the premise is correct, but loosening only the provider here wouldn't actually unblock proxy deployments, so I'm holding off.

The provider is faithfully mirroring the server's create gate. `CreateAIProviderRequest.Validate()` in `codersdk/aiproviders.go` rejects `type=bedrock` when `req.Settings.Bedrock == nil || !req.Settings.Bedrock.IsConfigured()` — and `IsConfigured()` only checks `region` + credentials, ignoring `base_url`, exactly like this check. The create handler (`coderd/ai_providers.go`) calls that same `Validate()`, so a base-URL-only Bedrock provider (AWS SDK default credential chain, no region) currently returns an HTTP 400 from Coder regardless of what Terraform does.

The base-URL-aware `IsBedrockConfigured(baseURL, settings)` does exist, but it's only used by legacy env-var seeding (`cli/server.go`) and migration detection (`coderd/ai_providers_migrate.go`) — never in the create/update API validation path.

So if I switch this check to be base-URL-aware in isolation, the apply still fails — just with a server-side 400 instead of a clean client-side diagnostic, which is worse UX. This is an inconsistency in Coder itself between the Bedrock docs and `CreateAIProviderRequest.Validate()`. The right fix is upstream: have `CreateAIProviderRequest.Validate()` use `IsBedrockConfigured(baseURL, ...)` (or correct the docs). Once the API accepts base-URL-only Bedrock, I'll mirror it here.

ethanndickson:
@codex review

ethanndickson:
@codex review

re: 

> Accept Bedrock custom endpoints with default credentials
> Honor full Bedrock base URLs without a region


Verified this against coder/coder at current main — the premise is correct, but loosening only the provider here wouldn't actually unblock proxy deployments, so I'm holding off.

The provider is faithfully mirroring the server's create gate. CreateAIProviderRequest.Validate() in codersdk/aiproviders.go rejects type=bedrock when req.Settings.Bedrock == nil || !req.Settings.Bedrock.IsConfigured() — and IsConfigured() only checks region + credentials, ignoring base_url, exactly like this check. The create handler (coderd/ai_providers.go) calls that same Validate(), so a base-URL-only Bedrock provider (AWS SDK default credential chain, no region) currently returns an HTTP 400 from Coder regardless of what Terraform does.

The base-URL-aware IsBedrockConfigured(baseURL, settings) does exist, but it's only used by legacy env-var seeding (cli/server.go) and migration detection (coderd/ai_providers_migrate.go) — never in the create/update API validation path.

So if I switch this check to be base-URL-aware in isolation, the apply still fails — just with a server-side 400 instead of a clean client-side diagnostic, which is worse UX. This is an inconsistency in Coder itself between the Bedrock docs and CreateAIProviderRequest.Validate(). The right fix is upstream: have CreateAIProviderRequest.Validate() use IsBedrockConfigured(baseURL, ...) (or correct the docs). Once the API accepts base-URL-only Bedrock, I'll mirror it here

ethanndickson:
* **#378** <a href="https://app.graphite.com/github/pr/coder/terraform-provider-coderd/378?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#374** <a href="https://app.graphite.com/github/pr/coder/terraform-provider-coderd/374?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#372** <a href="https://app.graphite.com/github/pr/coder/terraform-provider-coderd/372?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#371** <a href="https://app.graphite.com/github/pr/coder/terraform-provider-coderd/371?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#369** <a href="https://app.graphite.com/github/pr/coder/terraform-provider-coderd/369?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#368** <a href="https://app.graphite.com/github/pr/coder/terraform-provider-coderd/368?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/terraform-provider-coderd/368?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#370** <a href="https://app.graphite.com/github/pr/coder/terraform-provider-coderd/370?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

ethanndickson:
@codex review

ethanndickson:
@codex review

re: 

> Accept Bedrock custom endpoints with default credentials
> Honor full Bedrock base URLs without a region


Verified this against coder/coder at current main — the premise is correct, but loosening only the provider here wouldn't actually unblock proxy deployments, so I'm holding off.

The provider is faithfully mirroring the server's create gate. CreateAIProviderRequest.Validate() in codersdk/aiproviders.go rejects type=bedrock when req.Settings.Bedrock == nil || !req.Settings.Bedrock.IsConfigured() — and IsConfigured() only checks region + credentials, ignoring base_url, exactly like this check. The create handler (coderd/ai_providers.go) calls that same Validate(), so a base-URL-only Bedrock provider (AWS SDK default credential chain, no region) currently returns an HTTP 400 from Coder regardless of what Terraform does.

The base-URL-aware IsBedrockConfigured(baseURL, settings) does exist, but it's only used by legacy env-var seeding (cli/server.go) and migration detection (coderd/ai_providers_migrate.go) — never in the create/update API validation path.

So if I switch this check to be base-URL-aware in isolation, the apply still fails — just with a server-side 400 instead of a clean client-side diagnostic, which is worse UX. This is an inconsistency in Coder itself between the Bedrock docs and CreateAIProviderRequest.Validate(). The right fix is upstream: have CreateAIProviderRequest.Validate() use IsBedrockConfigured(baseURL, ...) (or correct the docs). Once the API accepts base-URL-only Bedrock, I'll mirror it here

ethanndickson:
@codex review

re: 

> Accept Bedrock custom endpoints with default credentials
> Honor full Bedrock base URLs without a region


Verified this against coder/coder at current main — the premise is correct, but loosening only the provider here wouldn't actually unblock proxy deployments, so I'm holding off.

The provider is faithfully mirroring the server's create gate. CreateAIProviderRequest.Validate() in codersdk/aiproviders.go rejects type=bedrock when req.Settings.Bedrock == nil || !req.Settings.Bedrock.IsConfigured() — and IsConfigured() only checks region + credentials, ignoring base_url, exactly like this check. The create handler (coderd/ai_providers.go) calls that same Validate(), so a base-URL-only Bedrock provider (AWS SDK default credential chain, no region) currently returns an HTTP 400 from Coder regardless of what Terraform does.

The base-URL-aware IsBedrockConfigured(baseURL, settings) does exist, but it's only used by legacy env-var seeding (cli/server.go) and migration detection (coderd/ai_providers_migrate.go) — never in the create/update API validation path.

So if I switch this check to be base-URL-aware in isolation, the apply still fails — just with a server-side 400 instead of a clean client-side diagnostic, which is worse UX. This is an inconsistency in Coder itself between the Bedrock docs and CreateAIProviderRequest.Validate(). The right fix is upstream: have CreateAIProviderRequest.Validate() use IsBedrockConfigured(baseURL, ...) (or correct the docs). Once the API accepts base-URL-only Bedrock, I'll mirror it here

internal/provider/ai_provider_resource.go
```diff
@@ -0,0 +1,654 @@
...
+		validateAIProviderBaseURL(resp.Diagnostics.AddAttributeError, path.Root("base_url"), baseURL)
+	}
+
+	if data.Type.IsUnknown() {
+		return
+	}
+	providerType := codersdk.AIProviderType(data.Type.ValueString())
+
+	if !data.APIKeyWO.IsNull() && !data.APIKeyWO.IsUnknown() && (providerType == codersdk.AIProviderTypeBedrock || providerType == codersdk.AIProviderTypeCopilot) {
+		resp.Diagnostics.AddAttributeError(path.Root("api_key_wo"), "Invalid Attribute Combination", fmt.Sprintf("`api_key_wo` must not be configured when `type` is `%s`.", providerType))
```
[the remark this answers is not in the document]
ethanndickson: Fixed. `ValidateConfig` now rejects `api_key_wo` whenever `settings.bedrock` is set (not just `type` `bedrock`/`copilot`), so an Anthropic+Bedrock config with `api_key_wo` now fails with a provider diagnostic instead of an apply-time HTTP 400. Covered by the `api key rejected for anthropic with bedrock settings` case.

internal/provider/ai_provider_resource.go
```diff
@@ -0,0 +1,616 @@
...
+					"bedrock": schema.SingleNestedAttribute{
+						MarkdownDescription: "AWS Bedrock settings. Valid only for `type = \"bedrock\"` or `type = \"anthropic\"`.",
+						Optional:            true,
+						Attributes: map[string]schema.Attribute{
+							"region": schema.StringAttribute{
+								MarkdownDescription: "AWS region for Bedrock. If omitted, derived from the canonical Bedrock `base_url` attribute.",
+								Optional:            true,
+								Computed:            true,
+								PlanModifiers: []planmodifier.String{
+									stringplanmodifier.UseStateForUnknown(),
```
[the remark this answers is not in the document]
ethanndickson: Fixed in the review commit. Removed `UseStateForUnknown()` from `settings.bedrock.region`, so an unpinned region re-derives from `base_url` instead of pinning stale state. When `base_url` drives the region it now plans as known-after-apply, eliminating the inconsistency error. Region derivation is anchored by the new `TestParseBedrockRegionFromBaseURL`.

internal/provider/ai_provider_resource.go
```diff
@@ -0,0 +1,616 @@
...
+	// untouched; dropping the bedrock block clears settings server-side.
+	credentialsChanged := credentialsVersionChanged(m.bedrock(), state.bedrock())
+	if m.bedrock() != nil || state.bedrock() != nil {
+		settings := m.sdkSettings(config, credentialsChanged, diags)
+		patch.Settings = &settings
+	}
+
+	// Only touch keys when the version changes. The whole key set is one
+	// key: send the new plaintext to rotate, or an empty list to clear.
+	if !m.APIKeyWOVersion.Equal(state.APIKeyWOVersion) {
```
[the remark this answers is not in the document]
ethanndickson: Fixed. `updateRequest` no longer sends an empty mutation list on a version change. A null `api_key_wo_version` now means unmanaged/preserve (no `APIKeys` sent), and bumping the version without a configured `api_key_wo` returns a `Missing API Key` diagnostic instead of silently clearing. Covered by `TestAIProviderUpdateRequestAPIKeyRotation` (version-removed and version-bumped-without-key cases).

internal/provider/ai_provider_resource.go
```diff
@@ -0,0 +1,616 @@
...
+	}
+}
+
+func bedrockCredentialsConfigured(b *AIProviderBedrockSettingsModel) bool {
+	return b != nil &&
+		!b.AccessKeyWO.IsNull() && !b.AccessKeyWO.IsUnknown() &&
+		!b.AccessKeySecretWO.IsNull() && !b.AccessKeySecretWO.IsUnknown()
+}
+
+func credentialsVersionChanged(a, b *AIProviderBedrockSettingsModel) bool {
```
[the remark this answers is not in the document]
ethanndickson: Fixed. `credentialsVersionChanged(plan, state)` now treats a null planned `credentials_wo_version` as unmanaged and returns false, so removing the field preserves server-side credentials instead of demanding them forever. Covered by `TestAIProviderUpdatePreservesBedrockCredentialsWhenVersionRemoved`.

internal/provider/ai_provider_resource_test.go
```diff
@@ -0,0 +1,442 @@
...
+	}
+}
+
+func TestAIProviderCreateRequestBedrockWithoutCredentials(t *testing.T) {
+	t.Parallel()
+
+	plan := AIProviderResourceModel{
+		Type:        types.StringValue(string(codersdk.AIProviderTypeBedrock)),
+		Name:        types.StringValue("aws-bedrock"),
+		DisplayName: types.StringUnknown(),
```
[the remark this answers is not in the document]
ethanndickson: Fixed. The test now deep-copies the nested Bedrock value (`configBedrock := *plan.Settings.Bedrock`; `config.Settings = &AIProviderSettingsModel{Bedrock: &configBedrock}`) so `config` gets a null region while `plan` keeps the unknown region, exercising the intended precondition rather than both-null.

internal/provider/ai_provider_resource_test.go
```diff
@@ -0,0 +1,442 @@
...
+						// PlanOnly expects an empty plan unless this is set.
+						ExpectNonEmptyPlan: true,
+					},
+				},
+			})
+		})
+	}
+}
+
+func TestAIProviderCreateRequestBedrockWithoutCredentials(t *testing.T) {
```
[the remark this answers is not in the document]
ethanndickson: Added. `TestAIProviderUpdateRejectsBedrockCredentialBumpWithoutCredentials` covers the `sdkSettings` credential-error path, and `TestAIProviderUpdatePreservesBedrockCredentialsWhenVersionRemoved` covers the preserve path, so the `credentialsVersionChanged` version comparison is now exercised.

internal/provider/ai_provider_resource.go
```diff
@@ -0,0 +1,616 @@
...
+		// Write-only and version values are never returned by the API;
+		// preserve the configured/state values.
+		APIKeyWO:        types.StringNull(),
+		APIKeyWOVersion: m.APIKeyWOVersion,
+		APIKeyMasked:    types.StringNull(),
+	}
+	if len(provider.APIKeys) > 0 {
+		out.APIKeyMasked = types.StringValue(provider.APIKeys[0].Masked)
+	}
+	if provider.Settings.Bedrock != nil {
```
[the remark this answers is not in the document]
ethanndickson: Fixed. `ValidateConfig` now rejects an empty `settings = {}` for non-Bedrock providers (`settings must include a bedrock block or be omitted`), avoiding the null-vs-empty perpetual diff. Covered by the `empty settings rejected for non-bedrock` case.

internal/provider/ai_provider_resource.go
```diff
@@ -0,0 +1,616 @@
...
+							"small_fast_model": schema.StringAttribute{
+								MarkdownDescription: "Small/fast Bedrock model identifier used for background tasks.",
+								Optional:            true,
+								Computed:            true,
+								PlanModifiers: []planmodifier.String{
+									stringplanmodifier.UseStateForUnknown(),
+								},
+							},
+							"access_key_wo": schema.StringAttribute{
+								MarkdownDescription: "AWS access key ID for Bedrock. See [Coder's Amazon Bedrock provider docs](https://coder.com/docs/@v2.34.3/ai-coder/ai-gateway/providers#amazon-bedrock).",
```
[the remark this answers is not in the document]
ethanndickson: Fixed. Unpinned to `https://coder.com/docs/ai-coder/ai-gateway/providers#amazon-bedrock` and regenerated docs.

internal/provider/ai_provider_resource_test.go
```diff
@@ -0,0 +1,442 @@
+package provider
```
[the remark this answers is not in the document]
ethanndickson: Added. `TestAIProviderResourceSchemaValidation` now covers `api_key_wo` rejection for `copilot`, `settings.bedrock` rejection for `openai`, and invalid `base_url`.

internal/provider/ai_provider_resource.go
```diff
@@ -0,0 +1,616 @@
...
+	// credentials, so omitting credential pointers leaves stored AWS keys
+	// untouched; dropping the bedrock block clears settings server-side.
+	credentialsChanged := credentialsVersionChanged(m.bedrock(), state.bedrock())
+	if m.bedrock() != nil || state.bedrock() != nil {
+		settings := m.sdkSettings(config, credentialsChanged, diags)
+		patch.Settings = &settings
+	}
+
+	// Only touch keys when the version changes. The whole key set is one
+	// key: send the new plaintext to rotate, or an empty list to clear.
```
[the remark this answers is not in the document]
ethanndickson: Fixed by removing the clear-by-omission path entirely rather than documenting it. A version bump without a key now errors, and a null version preserves the stored key, so there is no ambiguous empty-list-to-clear semantic left to document.

internal/provider/ai_provider_resource.go
```diff
@@ -0,0 +1,616 @@
...
+			"Unexpected Resource Configure Type",
+			fmt.Sprintf("Expected *CoderdProviderData, got: %T. Please report this issue to the provider developers.", req.ProviderData),
+		)
+		return
+	}
+	r.data = data
+}
+
+func (r *AIProviderResource) ValidateConfig(ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse) {
+	// The pointer-based model can't decode an unknown settings/bedrock object
```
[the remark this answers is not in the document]
ethanndickson: Trimmed the most verbose comments in the review commit (the `credentialsVersionChanged` doc, the API-key rotation block, and two `ValidateConfig` diagnostics). The two remaining 2-line comments you cite document non-obvious behavior (pointer model can't decode unknown nested objects; the Coder API rejects empty patches), so I kept them intentionally.

internal/provider/ai_provider_resource.go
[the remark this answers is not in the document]
ethanndickson: Fixed. Renamed to `credentialsVersionChanged(plan, state)` to make the directional contract explicit.

internal/provider/ai_provider_resource.go
```diff
@@ -0,0 +1,616 @@
...
+		resp.Diagnostics.AddAttributeError(path.Root("settings").AtName("bedrock"), "Invalid Attribute Combination", "`settings.bedrock` is only valid when `type` is `anthropic` or `bedrock`.")
+	}
+	accessSet := !bedrock.AccessKeyWO.IsNull() && !bedrock.AccessKeyWO.IsUnknown()
+	secretSet := !bedrock.AccessKeySecretWO.IsNull() && !bedrock.AccessKeySecretWO.IsUnknown()
+	if providerType == codersdk.AIProviderTypeBedrock {
+		if !baseURLKnown || bedrock.Region.IsUnknown() || bedrock.AccessKeyWO.IsUnknown() || bedrock.AccessKeySecretWO.IsUnknown() {
+			return
+		}
+		sdkSettings := codersdk.AIProviderBedrockSettings{
+			Region:         bedrockRegion(baseURL, bedrock.Region, bedrock.Region),
```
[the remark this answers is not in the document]
ethanndickson: Acknowledged, keeping as-is. In `ValidateConfig` there is no plan/config divergence, so tier 1 always wins and passing `bedrock.Region` for both args is intentional reuse of the shared helper. Minor and not worth a dedicated wrapper.

internal/provider/ai_provider_resource_test.go
```diff
@@ -0,0 +1,442 @@
...
+func testAIProviderTerraformVersionChecks() []tfversion.TerraformVersionCheck {
+	return []tfversion.TerraformVersionCheck{
+		tfversion.SkipBelow(tfversion.Version1_11_0),
+	}
+}
+
+func TestAccAIProviderResource(t *testing.T) {
+	t.Parallel()
+	if os.Getenv("TF_ACC") == "" {
+		t.Skip("Acceptance tests are disabled.")
```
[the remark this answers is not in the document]
ethanndickson: Acknowledged as a deferred note. Bedrock request-building logic is unit-tested; the remaining gap is server-integration acceptance coverage, which we are deferring unless we add Bedrock acceptance steps.

internal/provider/ai_provider_resource.go
```diff
@@ -0,0 +1,616 @@
...
+	if region := parseBedrockRegionFromBaseURL(baseURL); region != "" {
+		return region
+	}
+	if !planned.IsNull() && !planned.IsUnknown() {
+		return planned.ValueString()
+	}
+	return ""
+}
+
+func parseBedrockRegionFromBaseURL(baseURL string) string {
```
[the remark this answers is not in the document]
ethanndickson: Added. `TestParseBedrockRegionFromBaseURL` is a table test covering trailing slash, mixed-case host, surrounding whitespace, trailing path, query string, port, non-bedrock host, and empty input.

internal/provider/ai_provider_resource.go
```diff
@@ -0,0 +1,616 @@
...
+	match := bedrockCanonicalBaseURLRegex.FindStringSubmatch(strings.TrimSpace(baseURL))
+	if len(match) != 2 {
+		return ""
+	}
+	return strings.ToLower(match[1])
+}
+
+func validateAIProviderBaseURL(addError func(path.Path, string, string), attrPath path.Path, raw string) {
```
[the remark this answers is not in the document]
ethanndickson: Acknowledged as a deferred note. The SDK validation has the same gap and `base_url` is an admin-only endpoint field, so we are not adding userinfo hardening here.

internal/provider/ai_provider_resource.go
```diff
@@ -0,0 +1,616 @@
...
+
+	bedrock := data.bedrock()
+	if bedrock == nil {
+		if providerType == codersdk.AIProviderTypeBedrock {
+			resp.Diagnostics.AddAttributeError(path.Root("settings"), "Missing Bedrock Settings", "`type = \"bedrock\"` requires `settings.bedrock` with at least `region` or write-only AWS credentials.")
+		}
+		return
+	}
+
+	if providerType != codersdk.AIProviderTypeAnthropic && providerType != codersdk.AIProviderTypeBedrock {
```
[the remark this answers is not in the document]
ethanndickson: Acknowledged, intentional. This matches the SDK `Validate()` behavior: `type = anthropic` with `settings.bedrock` is allowed without the `IsConfigured()` completeness check, and the server decides usability. Won't-fix provider-side.

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

re: 

> Accept Bedrock custom endpoints with default credentials
> Honor full Bedrock base URLs without a region


Verified this against coder/coder at current main — the premise is correct, but loosening only the provider here wouldn't actually unblock proxy deployments, so I'm holding off.

The provider is faithfully mirroring the server's create gate. CreateAIProviderRequest.Validate() in codersdk/aiproviders.go rejects type=bedrock when req.Settings.Bedrock == nil || !req.Settings.Bedrock.IsConfigured() — and IsConfigured() only checks region + credentials, ignoring base_url, exactly like this check. The create handler (coderd/ai_providers.go) calls that same Validate(), so a base-URL-only Bedrock provider (AWS SDK default credential chain, no region) currently returns an HTTP 400 from Coder regardless of what Terraform does.

The base-URL-aware IsBedrockConfigured(baseURL, settings) does exist, but it's only used by legacy env-var seeding (cli/server.go) and migration detection (coderd/ai_providers_migrate.go) — never in the create/update API validation path.

So if I switch this check to be base-URL-aware in isolation, the apply still fails — just with a server-side 400 instead of a clean client-side diagnostic, which is worse UX. This is an inconsistency in Coder itself between the Bedrock docs and CreateAIProviderRequest.Validate(). The right fix is upstream: have CreateAIProviderRequest.Validate() use IsBedrockConfigured(baseURL, ...) (or correct the docs). Once the API accepts base-URL-only Bedrock, I'll mirror it here

docs/resources/experimental_ai_provider.md
```diff
@@ -0,0 +1,119 @@
...
+Experimental Coder AI provider configuration.
+
+`_wo` attributes are [write-only](https://developer.hashicorp.com/terraform/language/resources/ephemeral#write-only-arguments): their values are sent to Coder but never stored in Terraform state. This resource therefore requires Terraform 1.11 or later.
+
+For `type = "bedrock"`, omit `settings.bedrock.access_key_wo` and `settings.bedrock.access_key_secret_wo` to use the AWS SDK default credential chain as resolved by the Coder server process (IAM role, IRSA, environment variables, shared config, SSO, IMDS, and more). Set both together to use static IAM-user credentials.
+
+## Example Usage
+
+```terraform
+resource "coderd_experimental_ai_provider" "bedrock" {
```
matifali: I would suggest against naming it `experimental`<br>We shouldn't mention in description that it is experimental resource. <br><br>This would save us from deprecating this and later moving to a non-experimental resource. Without leaving a dead resource in the docs. <br><br>Removing or renaming a resource requires bumping the major provider version.

docs/resources/ai_provider.md
```diff
@@ -0,0 +1,119 @@
...
+- `base_url` (String) Absolute HTTP(S) base URL for the upstream provider endpoint.
+- `name` (String) Unique provider name. Must be lowercase alphanumeric words separated by hyphens.
+- `type` (String) AI provider type. Valid values are `openai`, `anthropic`, `azure`, `bedrock`, `google`, `openai-compat`, `openrouter`, `vercel`, and `copilot`.
+
+### Optional
+
+> **NOTE**: [Write-only arguments](https://developer.hashicorp.com/terraform/language/resources/ephemeral#write-only-arguments) are supported in Terraform 1.11 and later.
+
+- `api_key_wo` (String, Sensitive, [Write-only](https://developer.hashicorp.com/terraform/language/resources/ephemeral#write-only-arguments)) Plaintext API key for the provider. Not valid for `bedrock` or `copilot`, or when `settings.bedrock` is set. Bump `api_key_wo_version` to rotate it.
+- `api_key_wo_version` (Number) Version for the write-only API key. Required when `api_key_wo` is set; bump it whenever `api_key_wo` changes to rotate the stored key.
```
matifali, line 72: Do we really need `_wo` in the argument name? I think it's enough to mention that in the description.

matifali (commented):
I have concerns about the choice of the name for this resource.

ethanndickson:
Yeah honestly I agree, I think putting experimental in the title is a mistake. It'll be extremely painful for customers to migrate when we remove it.

docs/resources/ai_provider.md
matifali: NVM just read linked docs. This is what they show in the example.

docs/resources/experimental_ai_provider.md
ethanndickson: Agreed, but FWIW, Terraform actually does support moving resource state to a resource with a different name:
https://developer.hashicorp.com/terraform/plugin/framework/resources/state-move#use-cases
Users do need to add a `moved` block to their Terraform, however.

ethanndickson: Will bring it up with the team.

matifali: Nice. TIL i am ok if there is a good enough mechanism. <br>I am absing my arugment on [https://developer.hashicorp.com/terraform/plugin/best-practices/versioning#example-major-number-increments](<https://developer.hashicorp.com/terraform/plugin/best-practices/versioning#example-major-number-increments>)<br>where removing a resource counts as a major version bump. If we want to do that, it's fine.

ethanndickson: IMO if we were to do the rename, I'd probably skip the major version bump. We'd be promoting an experimental resource to stable - no breaking changes unless you're opting in to using the experimental resource(s)

matifali: Yes we can do that. I am just talking about the recommendations but we do not have to be strict given its an experimental resource we are removing.

docs/resources/ai_provider.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Required
+
+- `base_url` (String) Absolute HTTP(S) base URL for the upstream provider endpoint.
+- `name` (String) Unique provider name. Must be lowercase alphanumeric words separated by hyphens.
+- `type` (String) AI provider type. Valid values are `openai`, `anthropic`, `azure`, `bedrock`, `google`, `openai-compat`, `openrouter`, `vercel`, and `copilot`.
+
+### Optional
+
+> **NOTE**: [Write-only arguments](https://developer.hashicorp.com/terraform/language/resources/ephemeral#write-only-arguments) are supported in Terraform 1.11 and later.
```
matifali, line 69:
```suggestion
-> [Write-only arguments](https://developer.hashicorp.com/terraform/language/resources/ephemeral#write-only-arguments) are supported in Terraform 1.11 and later.
```
This could also turn into a Terraform registry-style note. 

Image
Preview from: https://registry.terraform.io/tools/doc-preview

ethanndickson: There's already one at the top, I'm going to get rid of it

ethanndickson: Nevermind, it's auto-generated and you can't get rid of or edit it.

ethanndickson: Incredible.

ethanndickson: I think the block at the top saying the whole resource requires 1.11 or later is worth keeping, so keeping both unfortunately.

matifali: Yes we can drop this block and keep the top only.

ethanndickson:
```
> **NOTE**: [Write-only arguments](https://developer.hashicorp.com/terraform/language/resources/ephemeral#write-only-arguments) are supported in Terraform 1.11 and later.
```

You can't get rid of this, it's added by the docs generator.

examples/resources/coderd_ai_provider/resource.tf
```diff
@@ -0,0 +1,31 @@
...
+    }
+  }
+}
+
+resource "coderd_ai_provider" "openai" {
+  type         = "openai"
+  name         = "openai"
+  display_name = "OpenAI"
+  enabled      = true
+  base_url     = "https://api.openai.com"
```
ethanndickson: Note to future self to investigate: does coderd handle omitting the `/v1` for openAI base urls? otherwise we should probably catch this during validation/plan

internal/provider/ai_provider_resource.go
```diff
@@ -0,0 +1,727 @@
...
+		"Experimental Resource",
+		"coderd_ai_provider is experimental. Changes are expected, and it is not recommended for production use.",
+	)
+}
+
+func (r *AIProviderResource) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		MarkdownDescription: "~> This resource is experimental. Changes are expected, and it is not recommended for production use.\n\n" +
+			"-> `_wo` attributes are [write-only](https://developer.hashicorp.com/terraform/language/resources/ephemeral#write-only-arguments): " +
+			"their values are sent to Coder but never stored in Terraform state. This resource therefore requires Terraform 1.11 or later.\n\n" +
```
johnstcn, line 87: TIL, this is cool

examples/resources/coderd_ai_provider/resource.tf
ethanndickson: Added a warning to the terraform logs when /v1 isn't present in the path for openai & openai-compat.

johnstcn: I checked on dev.coder.com and the openai(-compat) providers all have the `/v1` suffix.

johnstcn (approved):

ethanndickson:
Going to handle the new `role_arn` field (as introduced by AIGOV-371) in a follow-up PR.

internal/provider/ai_provider_resource.go
```diff
@@ -0,0 +1,738 @@
...
+}
+
+type AIProviderResourceModel struct {
+	ID              UUID                     `tfsdk:"id"`
+	Type            types.String             `tfsdk:"type"`
+	Name            types.String             `tfsdk:"name"`
+	DisplayName     types.String             `tfsdk:"display_name"`
+	Enabled         types.Bool               `tfsdk:"enabled"`
+	BaseURL         types.String             `tfsdk:"base_url"`
+	APIKeyWO        types.String             `tfsdk:"api_key_wo"`
```
dannykopping, line 51: Cool solution! I'd prefer we didn't couple the implementation (write-only secret) with the field name.

Let's be specific and call it `api_key_plaintext` or something?

internal/provider/ai_provider_resource.go
```diff
@@ -0,0 +1,738 @@
...
+				Optional:            true,
+				Computed:            true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"enabled": schema.BoolAttribute{
+				MarkdownDescription: "Whether this AI provider is enabled. Defaults to true.",
+				Optional:            true,
+				Computed:            true,
```
dannykopping, line 140: Howcome this is computed? It can be defined when creating a provider.

internal/provider/ai_provider_resource.go
```diff
@@ -0,0 +1,738 @@
...
+								Optional:            true,
+								Computed:            true,
+								PlanModifiers: []planmodifier.String{
+									bedrockRegionPlanModifier{},
+								},
+							},
+							"model": schema.StringAttribute{
+								MarkdownDescription: "Primary Bedrock model identifier.",
+								Optional:            true,
+								Computed:            true,
```
dannykopping, line 184: Not computed; it's configurable.

internal/provider/ai_provider_resource.go
```diff
@@ -0,0 +1,738 @@
...
+								PlanModifiers: []planmodifier.String{
+									stringplanmodifier.UseStateForUnknown(),
+								},
+							},
+							"small_fast_model": schema.StringAttribute{
+								MarkdownDescription: "Small/fast Bedrock model identifier used for background tasks.",
+								Optional:            true,
+								Computed:            true,
```
dannykopping, line 192: Ditto.

internal/provider/ai_provider_resource.go
```diff
@@ -0,0 +1,738 @@
...
+	settings := codersdk.AIProviderBedrockSettings{
+		Region: bedrockRegion(m.BaseURL.ValueString(), cfgBedrock.Region, bedrock.Region),
+	}
+	if !cfgBedrock.AccessKeyWO.IsNull() && !cfgBedrock.AccessKeyWO.IsUnknown() {
+		accessKey := cfgBedrock.AccessKeyWO.ValueString()
+		settings.AccessKey = &accessKey
+	}
+	if !cfgBedrock.AccessKeySecretWO.IsNull() && !cfgBedrock.AccessKeySecretWO.IsUnknown() {
+		accessKeySecret := cfgBedrock.AccessKeySecretWO.ValueString()
+		settings.AccessKeySecret = &accessKeySecret
```
dannykopping: Nit: might be good to add an abstraction which handles this. Seeing a lot of repeated patterns here around this write-only value.

internal/provider/ai_provider_resource.go
```diff
@@ -0,0 +1,738 @@
...
+		DisplayName: types.StringValue(provider.DisplayName),
+		Enabled:     types.BoolValue(provider.Enabled),
+		BaseURL:     types.StringValue(provider.BaseURL),
+		CreatedAt:   types.Int64Value(provider.CreatedAt.Unix()),
+		UpdatedAt:   types.Int64Value(provider.UpdatedAt.Unix()),
+		// Write-only and version values are never returned by the API;
+		// preserve the configured/state values.
+		APIKeyWO:        types.StringNull(),
+		APIKeyWOVersion: m.APIKeyWOVersion,
+		APIKeyMasked:    types.StringNull(),
```
dannykopping: This isn't write-only.

dannykopping (commented):
Looking good, thanks a mil for doing this @ethanndickson!

Have you tested this against a local instance?

internal/provider/ai_provider_resource.go
ethanndickson: Suffxing with `_wo` is the recommended style for write only arguments.
https://developer.hashicorp.com/terraform/plugin/best-practices/naming#write-only-argument-names

internal/provider/ai_provider_resource.go
ethanndickson: Computed just means it can be set even if it's omitted in the config, so any value that defaults must be `Computed`.

internal/provider/ai_provider_resource.go
ethanndickson: It's optional and computed, since they have default values. (`global.anthropic.claude-sonnet-4-5-20250929-v1:0`, `global.anthropic.claude-haiku-4-5-20251001-v1:0`)

internal/provider/ai_provider_resource.go
ethanndickson: See above

internal/provider/ai_provider_resource.go
ethanndickson: You could argue it's less important since we dont have a non `wo` alternative, but I think that makes api_key_version read weird, why is the api key versioned? I'd argue `api_key_wo_version` makes it more clear.

internal/provider/ai_provider_resource.go
ethanndickson: Will fix the ordering so those two come last.

ethanndickson:
> Have you tested this against a local instance?

Yep, and #371 adds an automated integration test that runs this resource alongside the coder agents model resource.

internal/provider/ai_provider_resource.go
dannykopping: Doh, should've done some more reading on WO args. Thanks 👍

dannykopping (approved):

matifali (approved):

ethanndickson:
### Merge activity

* **Jul 1, 5:55 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/coder/terraform-provider-coderd/368).
* **Jul 1, 5:56 AM UTC**: [Graphite](https://app.graphite.com/github/pr/coder/terraform-provider-coderd/368) rebased this pull request as part of a merge.
* **Jul 1, 5:56 AM UTC**: @ethanndickson merged this pull request with [Graphite](https://app.graphite.com/github/pr/coder/terraform-provider-coderd/368).

PR #374, opened 2026-06-26
test: migrate Docker orchestration to testcontainers-go

ethanndickson:
Replaces the hand-rolled Docker SDK container lifecycle in `integration/integration.go` with `testcontainers-go`. `StartCoder`'s signature and all its options are unchanged, so the acceptance and integration test callsites are untouched.

testcontainers handles image pulls, the mapped host port, readiness waiting, and cleanup, which lets us delete the `pullImage` and `randomPort` helpers and drop `docker/docker` as a direct dependency.

Closes PLAT-367.

PR #372, opened 2026-06-26
feat(coderd_ai_provider): add Bedrock role_arn for STS assume-role

ethanndickson:
## Background

A customer running AWS Bedrock needs runtime IAM role assumption across multiple AWS accounts so usage bills to the correct account: a pod starts with its ambient AWS identity (IRSA / EKS Pod Identity / instance profile), then assumes a downstream role in another account via STS before calling Bedrock, with temp credentials cached and rotated by the AWS SDK. Static keys are not acceptable, and the base identity alone is insufficient since it's shared across providers.

This gap ([AIGOV-371](https://linear.app/codercom/issue/AIGOV-371/support-dynamic-bedrock-assumerole-across-aws-accounts-for-ai-gateway)) was fixed server-side in coder/coder#26527 (backend: assume a configured role ARN via STS before calling Bedrock; whether the role is same- or cross-account is purely a matter of its trust policy) and coder/coder#26578 (web UI follow-up). The only piece reaching this provider is one new `role_arn` field on `codersdk.AIProviderBedrockSettings`, plus server-side ARN validation. Everything else lives in `aibridge` / `coderd` / `cli` / UI, which this provider doesn't import.

## What this PR does

Adds an optional `role_arn` to `coderd_ai_provider`'s `settings.bedrock` block so the assume-role workflow is configurable via Terraform, not only the UI. When set, the gateway uses its base identity to assume that IAM role via STS and signs Bedrock requests with the resulting temporary credentials; the role can live in the same account or a different one, depending entirely on its trust policy. A deployment that needs several roles configures several providers. Omitting it preserves today's behavior. Also updates the example and regenerates `docs/`.

## SDK bump

`role_arn` isn't in any tagged `codersdk` release yet (latest v2.34.3 predates it), so `go.mod` is pinned to a pseudo-version of the coder/coder#26527 merge commit; re-pin to a real tag once one ships.

Closes CODAGT-607. Refs AIGOV-371. Refs coder/coder#26527, coder/coder#26578.

PR #370, opened 2026-06-24
ci: run acceptance tests against an external Postgres sidecar

ethanndickson:
> **Stack** — base PR; #368 (experimental AI provider resource) builds on this.

## Problem

The Terraform acceptance matrix was failing intermittently across all lanes with `coder failed to become ready in time`.

`integration.StartCoder` boots `ghcr.io/coder/coder` without `CODER_PG_CONNECTION_URL`, so Coder falls back to its embedded PostgreSQL. The image doesn't bundle the Postgres binary, so each startup downloads the embedded-postgres jar from Maven Central. GitHub runners' shared egress IPs get rate-limited by Cloudflare, and a single non-200 crashes Coder before it binds — reddening the lane.

## Fix

Give Coder a real PostgreSQL instead of the embedded one. Setting `CODER_PG_CONNECTION_URL` bypasses the Maven download entirely. Per test, `integration.StartCoder` now starts a Postgres sidecar on a user-defined Docker network (aliased `postgres`), wires Coder onto it, and points `CODER_PG_CONNECTION_URL` at it. No readiness wait needed — Coder retries its DB connection for ~30s, covering sidecar boot.

The image is `us-docker.pkg.dev/coder-v2-images-public/public/postgres:17`, the public mirror coder/coder uses in its own tests, which avoids Docker Hub's anonymous pull rate limit.

This is environment-level, so it also covers the version-pinned back-compat tests. CI wall time is essentially unchanged (~206s/lane): the Postgres pull and boot costs about what the Maven download it replaces did.

## Also: raise the Terraform support floor to 1.5

Terraform 1.0–1.4 are EOL. This drops them from the CI matrix (now `1.5.*`–`1.14.*`, so we also test the latest releases), sets the README floor to `>= 1.5`, and removes the `template_resource_test.go` skip guard for a Terraform 1.0 panic.

PR #369, opened 2026-06-23
docs: add AGENTS.md contributor guide for AI agents

ethanndickson:
Adds an `AGENTS.md` contributor guide at the repo root to give AI coding agents the high-signal context they need to work in this provider safely.

The document captures the provider's framework idioms (terraform-plugin-framework, not SDKv2), the resource file layout, essential `make` commands, and — most importantly — the project-specific patterns and anti-patterns distilled from past fixes (e.g. handling unknown values at plan/validate time, refreshing cached entitlements after a license apply, not stripping cty sensitivity marks in plan modifiers, and avoiding `defer` inside retry loops). Each anti-pattern references the PR/issue it was learned from.

Contents were verified against the current tree: `Makefile` targets, the `util.go` helpers, `provider.go` entitlement plumbing, the CI Terraform `1.0`–`1.9` acceptance matrix, and the golangci `paralleltest` config.

PR #378, opened 2026-06-30
feat: add experimental coderd_default_agents_model resource

ethanndickson:
> **TL;DR** — Coder enforces exactly one default Agents chat model and flips that flag server-side, so a per-model `is_default` boolean can't be represented in Terraform without perpetual plan noise or "inconsistent result after apply." #371 removed `is_default` from `coderd_agents_model`; this PR adds the replacement — a small `coderd_default_agents_model` pointer resource that names which model is the default. This is the standard Terraform pattern for a server-enforced "exactly one of N" default.

> **Stack** — builds on #371 (`coderd_agents_model`), which removed `is_default` from the member resource in anticipation of this one. Review/merge the downstack PRs first.

Relates to CODAGT-607

## Summary

Adds `coderd_default_agents_model`, a singleton pointer resource that selects which `coderd_agents_model` is the deployment-wide default chat model for Coder Agents:

```hcl
resource "coderd_agents_model" "sonnet" {
  ai_provider_id = coderd_ai_provider.anthropic.id
  model          = "claude-3-5-sonnet-20241022"
  context_limit  = 200000
}

resource "coderd_default_agents_model" "this" {
  model_id = coderd_agents_model.sonnet.id
}
```

This completes the redesign started in #371. `is_default` was removed from `coderd_agents_model` there because Coder enforces exactly one default via a partial-unique index and mutates the flag out-of-band; a per-member `Optional + Computed` boolean either plans as perpetual `(known after apply)` noise or, if pinned, trips "Provider produced inconsistent result after apply" when a sibling is set default in the same apply and demotes this model after the plan is frozen. Moving the choice to its own resource removes both failure modes by construction: there's no per-member computed flag to go unknown, and nothing for a sibling to mutate underneath it.

## Why a separate pointer resource

This is the established Terraform pattern for "exactly one of N is the server-enforced default" — model the pointer as its own resource, never a per-member flag. Three mainstream providers do exactly this:

| Provider | Resource | Stack | Delete semantics |
|----------|----------|-------|------------------|
| AWS | `aws_ec2_default_credit_specification` | terraform-plugin-framework (our stack) | No-op — can't "unset", only repoint |
| AWS | `aws_ssm_default_patch_baseline` | SDKv2 | Restores the AWS-owned built-in default |
| GitHub | `github_branch_default` | SDKv2 | No-op, drops from state |

The closest analog is `aws_ec2_default_credit_specification`: same framework stack, and almost exactly our shape — a chosen value plus a scope, Create/Update both call the same "set default" API, and Delete is a deliberate no-op because AWS always keeps some default per scope and you can't reach "zero defaults." That's precisely Coder's constraint (the partial-unique index plus `ensureDefaultChatModelConfig` force-promotion). HashiCorp's `resource` package docs also explicitly sanction the create-only / no-meaningful-delete shape for one-time operations that need tracking, so a singleton pointer is a documented, intended use.

The counter-precedent is instructive: `aws_launch_template.default_version` keeps an `Optional + Computed` flag on the member and tolerates the churn — but only because launch-template versions are not a mutually-exclusive singleton across siblings, so it never hits the "sibling demoted out-of-band during the same apply" inconsistency. That structural difference is exactly why they can keep the per-member flag and we can't.

## Behavior

- **Create / Update** issue a single-field PATCH: `UpdateChatModelConfig(model_id, { is_default: true })`. The server's update handler is a read-modify-write merge that only overrides fields the request supplies, so a body of just `{"is_default": true}` preserves everything else, and the demote-of-previous + promote-of-new happens atomically in one transaction. This is an in-place update (like `aws_ec2_default_credit_specification`), not a replace.
- **Read** lists model configs and reflects whichever one the server currently reports as default into `model_id`; if no models exist it removes the resource from state.
- **Delete** is a no-op that just drops the resource from state. Coder always keeps exactly one model marked default and force-promotes a replacement when the current default is removed, so there's nothing to clear server-side — destroying the resource only stops Terraform from managing which model is the default. This is documented on the resource.
- **Import** takes the `coderd_agents_model` UUID and passes it through to `model_id`; Read then reconciles to the server's actual default, so a stale ID self-corrects on the next plan (mirrors `aws_ebs_default_kms_key`, which imports by the value it points at).

`id` is a computed constant (`default`) since the default is a global singleton with no scope key.

## Notes

The chat-model endpoints are experimental (`/api/experimental`), so the resource is marked experimental via a warning callout in its docs, matching `coderd_agents_model`. Ships with generated docs/examples, unit + acceptance tests, and an integration-test resource that points the default at a non-first model to prove it overrides the server's automatic election end-to-end.


ethanndickson:
* **#378** <a href="https://app.graphite.com/github/pr/coder/terraform-provider-coderd/378?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/terraform-provider-coderd/378?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#374** <a href="https://app.graphite.com/github/pr/coder/terraform-provider-coderd/374?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#372** <a href="https://app.graphite.com/github/pr/coder/terraform-provider-coderd/372?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#371** <a href="https://app.graphite.com/github/pr/coder/terraform-provider-coderd/371?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#369** <a href="https://app.graphite.com/github/pr/coder/terraform-provider-coderd/369?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#368** <a href="https://app.graphite.com/github/pr/coder/terraform-provider-coderd/368?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#370** <a href="https://app.graphite.com/github/pr/coder/terraform-provider-coderd/370?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

ethanndickson:
@codex review

internal/provider/default_agents_model_resource_test.go
```diff
@@ -0,0 +1,284 @@
...
+	require.Equal(t, defaultAgentsModelID, state.ID.ValueString())
+	require.Equal(t, id, state.ModelID.ValueUUID())
+	require.Equal(t, id.String(), state.ModelID.ValueString())
+}
+
+// TestDefaultAgentsModelResourceValidationDefersUnknownConfig proves the
+// validate walk succeeds when the Required model_id is unknown (sourced from a
+// no-default variable), the spot the #305 family of bugs surfaced. UUIDType
+// defers on unknown today; this guards against a future ValidateConfig or
+// validator that forgets the unknown guard.
```
johnstcn: Reword to use plainer language please!

internal/provider/default_agents_model_resource_test.go
```diff
@@ -0,0 +1,284 @@
...
+				ExpectNonEmptyPlan: true,
+			},
+		},
+	})
+}
+
+// TestAccDefaultAgentsModelResource exercises the full Terraform-managed
+// lifecycle: selecting a default, re-pointing it to a sibling model (which Coder
+// atomically demotes the previous default for), a steady-state no-drift re-plan,
+// and import.
```
johnstcn: re-word: given X, when Y, then Z.

internal/provider/default_agents_model_resource.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
+}
+
+func (r *DefaultAgentsModelResource) Delete(ctx context.Context, req resource.DeleteRequest, resp *resource.DeleteResponse) {
+	// No-op. Coder always keeps exactly one model marked as default and
+	// force-promotes a replacement when the current default is removed, so there
+	// is nothing to clear server-side. Removing this resource only stops
+	// Terraform from managing which model is default; the framework drops it from
+	// state automatically.
```
johnstcn: Comment re-states what the below log already says (also mentioned in the provider schema). We can probably drop it.

johnstcn (approved):
Nice work. Just some non-blocking nits about the comments. My general assumption is that when this passes the integration tests it's good to go, although a manual smoke-test is probably not a bad idea.

ethanndickson:
### Merge activity

* **Jul 1, 5:55 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/coder/terraform-provider-coderd/378).
* **Jul 1, 6:01 AM UTC**: [Graphite](https://app.graphite.com/github/pr/coder/terraform-provider-coderd/378) couldn't merge this pull request because a downstack PR #372 failed to merge.
* **Jul 1, 6:17 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/coder/terraform-provider-coderd/378).
* **Jul 1, 6:21 AM UTC**: [Graphite](https://app.graphite.com/github/pr/coder/terraform-provider-coderd/378) rebased this pull request as part of a merge.
* **Jul 1, 6:21 AM UTC**: @ethanndickson merged this pull request with [Graphite](https://app.graphite.com/github/pr/coder/terraform-provider-coderd/378).

PR #381, opened 2026-07-06
chore: bump coder/coder SDK and adapt to chat model config changes

ethanndickson:
Bumps `github.com/coder/coder/v2` so the next PR in this stack can use the Bedrock external ID from coder/coder#26869.

The bump also pulls in coder/coder#26877, which removed the `provider` field from chat model configs and made `ai_provider_id` required. The `coderd_agents_model` resource keeps its `provider_type` attribute, but now derives it by looking up the linked AI provider, which is upstream's stated migration path.

This stays compatible with older Coder servers: the provider never sent a `provider` field in requests, and the lookup used to derive `provider_type` hits the AI provider endpoint that predates this change. The response decode is also version-agnostic — the removed `provider` column is simply an unknown field that Go drops, and `ai_provider_id` decodes into the non-pointer `uuid.UUID` identically on both old and new servers.

Refs coder/coder#26877.

Closes CODAGT-753

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.26.4
 
 require (
 	cdr.dev/slog/v3 v3.1.0
-	github.com/coder/coder/v2 v2.34.0-rc.0.0.20260624160327-8bf6f4301614
+	github.com/coder/coder/v2 v2.34.0-rc.0.0.20260701204415-db7f4438b4ad
 	github.com/coder/retry v1.5.1
 	github.com/coder/serpent v0.15.0
 	github.com/coder/websocket v1.8.15
@@ -180,7 +180,7 @@ require (
 	github.com/tinylib/msgp v1.2.5 // indirect
 	github.com/tklauser/go-sysconf v0.3.16 // indirect
 	github.com/tklauser/numcpus v0.11.0 // indirect
-	github.com/valyala/fasthttp v1.71.0 // indirect
+	github.com/valyala/fasthttp v1.72.0 // indirect
 	github.com/vmihailenco/msgpack v4.0.4+incompatible // indirect
 	github.com/vmihailenco/msgpack/v5 v5.4.1 // indirect
 	github.com/vmihailenco/tagparser/v2 v2.0.0 // indirect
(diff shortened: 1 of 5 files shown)


ethanndickson:
* **#382** <a href="https://app.graphite.com/github/pr/coder/terraform-provider-coderd/382?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#381** <a href="https://app.graphite.com/github/pr/coder/terraform-provider-coderd/381?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/terraform-provider-coderd/381?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ethanndickson:
@codex review

ethanndickson:
@codex review

re: Handle legacy configs with missing AI provider IDs

Good thought, and the semantics are right — a null ai_provider_id does imply the config was soft-deleted (including the cascade when its AI provider is deleted, via DeleteChatModelConfigsByAIProviderID). But that deletion is already detected one layer up: soft-deleted rows are filtered out of ListChatModelConfigs (cmc.deleted = FALSE), so in Read a deleted config simply isn't in the returned list, and the existing fall-through handles it — AddWarning("...not found. Marking as deleted.") + RemoveResource(ctx).

So by the time lookupProviderType runs, the config either came from that list (Read) or is the active row we just wrote (Create/Update) — always non-null ai_provider_id. A null-triggered RemoveResource here would be dead code guarding a state we can't observe, since list-membership already covers the deletion case.

johnstcn (approved):

ethanndickson:
### Merge activity

* **Jul 8, 7:58 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/coder/terraform-provider-coderd/381).
* **Jul 8, 7:59 AM UTC**: @ethanndickson merged this pull request with [Graphite](https://app.graphite.com/github/pr/coder/terraform-provider-coderd/381).

PR #382, opened 2026-07-06
feat(coderd_ai_provider): expose server-generated Bedrock external_id

ethanndickson:
coder/coder#26869 added a server-generated STS external ID to Bedrock AI provider settings: when `role_arn` is set, the server generates an ID and sends it on every AssumeRole call, so operators can require it in the role's trust policy via an `sts:ExternalId` condition (the standard confused-deputy mitigation). This exposes it on `coderd_ai_provider` as a read-only `settings.bedrock.external_id` attribute, so the trust policy can reference it in the same Terraform config that manages the provider.

Design notes:

- The value is never sent outbound (the server rejects client-supplied values on create and non-matching values on update), only read back into state; empty responses from servers predating #26869 map to null, so the attribute works against both old and new coderd with no version gating.
- A custom plan modifier preserves a stored ID (the server carries it forward even if `role_arn` is later cleared), plans unknown while `role_arn` is set or unknown, and pins null otherwise. `UseStateForUnknown` would be wrong here per the "computed value derived from mutable config" anti-pattern: the empty-to-set `role_arn` transition would produce "inconsistent result after apply".

Builds on #381 (SDK bump). Follows up #372 (`role_arn`).

Closes AIGOV-504

### DIFF
--- a/docs/resources/ai_provider.md
+++ b/docs/resources/ai_provider.md
@@ -105,6 +105,10 @@ Optional:
 - `role_arn` (String) ARN of an AWS IAM role to assume via STS before calling Bedrock. The base identity (the AWS SDK default credential chain or the static credentials) signs the AssumeRole call, and the temporary credentials sign Bedrock requests. Omit to call Bedrock with the base identity directly. Requires Coder v2.35.0 or later.
 - `small_fast_model` (String) Small/fast Bedrock model identifier used for background tasks.
 
+Read-Only:
+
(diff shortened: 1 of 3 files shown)


ethanndickson:
* **#382** <a href="https://app.graphite.com/github/pr/coder/terraform-provider-coderd/382?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/terraform-provider-coderd/382?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#381** <a href="https://app.graphite.com/github/pr/coder/terraform-provider-coderd/381?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

ethanndickson:
@codex review

re: Preserve null external IDs from unsupported servers

Added a test case to prove this is a non-issue

evgeniy-scherbina (approved):

PR #383, opened 2026-07-06
feat(coderd_template): make versions optional

BobbyHo:
## Why

Some teams are moving template version rollout out of Terraform and into a
custom CI pipeline that calls `coder templates push` directly — this gives
them more control over how new versions are rolled out (e.g. pushing
inactive versions for testing, activating on a separate schedule, deploying
more than once a day). They still want Terraform to manage the template's
other settings: ACL, dormancy TTLs, autostop requirements, etc.

That split wasn't possible before this change, because `versions` on
`coderd_template` was `Required` and had to contain at least one element.
Every `coderd_template` resource was therefore forced to "own" at least one
version, which caused two problems:

1. There was no way to configure the resource at all without declaring a
   version — you couldn't manage settings-only.
2. Even with a throwaway/bootstrap version as a workaround, `Read()` would
   refresh that version's `active` flag from the API on every plan. Once the
   external pipeline activated a different version, the next `terraform
   apply` — even one only touching a TTL — would see that diff and call
   `UpdateActiveTemplateVersion` to reactivate Terraform's own (now stale)
   version, silently reverting whatever the pipeline had just rolled out.

## What changed

- **`versions` is now `Optional`** (dropped `Required` and the
  `listvalidator.SizeAtLeast(1)` validator on the schema attribute). Null or
  omitted `versions` now means "Terraform does not create, update, or read
  any template version — it only manages the template's other settings."
  This mirrors the existing `acl` attribute on this same resource, where
  `null` already means "don't manage this."
- **`Create()` now explicitly requires at least one version (with exactly
  one active)** before doing anything else. This can't be relaxed away:
  Coderd's `POST /templates` API itself requires a version ID to create a
  template, so `versions` can only be optional for templates that already
  exist (created previously by Terraform, or by `terraform import`).
- **Moved the "must have an active version" check** from the `versions`
  plan modifier into `Create()`. It used to infer "this is a brand-new
  resource" from "there's no prior private state yet," but that heuristic
  breaks for a `terraform import`ed template with `versions` omitted (no
  private state is ever written for an imported resource, since `Create()`
  never runs for it) — it would have raised a false "must be active" error
  on the very first plan after import. `Create()` is unambiguous: it's only
  ever invoked by Terraform Core for a genuine create.
- **The plan modifier now short-circuits on a null/unknown `versions`
  config**, before doing any reconciliation. Without this, rebuilding the
  planned list from zero elements would have produced a non-null empty list
  (`[]`) instead of preserving `null`, quietly turning "I didn't set this"
  into "I set it to empty."
- **`Read()` and `Update()` needed no changes.** Both already iterate over
  `Versions` with a plain Go `for range`, which is a no-op on a nil/empty
  slice — so "leave version management alone, just reconcile settings" was
  already their behavior once they could ever see an empty list.

## How it works

With `versions` left `null`, `Read()` never asks Coderd which version is
active, because it only loops over versions Terraform is already tracking
(zero, in this case) — there is nothing to compare against config, so there
is no drift to react to. Whatever the external pipeline does to the
template's active version is invisible to (and left untouched by)
Terraform. `Update()` only ever pushes/activates a version when
`Versions` is non-empty, so it never touches version state either.

## Test plan
- [x] `go build ./...` / `go vet ./...` clean
- [x] New `TestAccTemplateResourceOptionalVersions` (acceptance, against a
      real Coderd via testcontainers):
  - `CreateWithoutVersionsErrors` — creating a brand-new template with no
    versions fails with a clear error
  - `SettingsOnlyManagementAfterDroppingVersions` — create with Terraform,
    drop `versions` to `null`, have an external "pipeline" push and
    activate a new version directly against the API, then have Terraform
    update a setting — verifies Terraform never touches versions and the
    pipeline's active version survives
- [x] Full pre-existing `TestAccTemplateResource` suite reran — no
      regressions
- [x] Docs regenerated via `tfplugindocs generate`

## Manual Test

In addition to the automated tests above, I ran this end-to-end against a real
local Coder deployment with the real Terraform CLI (not the acceptance-test
harness), to confirm the actual customer workflow from the linked issue.

<details>
<summary><b>Setup: build the provider, point Terraform at it, start a local Coder</b></summary>

```shell
# 1. Build and install the modified provider
cd terraform-provider-coderd
go install .
# -> installs $(go env GOPATH)/bin/terraform-provider-coderd

# 2. Point Terraform at the local build (project-scoped, doesn't touch ~/.terraformrc)
cat > /tmp/dev.tfrc <<'EOF'
provider_installation {
  dev_overrides {
    "coder/coderd" = "/Users/<you>/go/bin"
  }
  direct {}
}
EOF
export TF_CLI_CONFIG_FILE=/tmp/dev.tfrc
# With dev_overrides active, skip `terraform init` for this provider.

# 3. Start a disposable local Coder deployment
docker network create coder-manual-test
docker run -d --name coder-pg --network coder-manual-test \
  -e POSTGRES_USER=coder -e POSTGRES_PASSWORD=coder -e POSTGRES_DB=coder postgres:17
docker run -d --name coder-server --network coder-manual-test -p 3000:3000 \
  -e CODER_PG_CONNECTION_URL="postgres://coder:coder@coder-pg/coder?sslmode=disable" \
  -e CODER_ACCESS_URL="http://localhost:3000" -e CODER_ADDRESS="0.0.0.0:3000" \
  ghcr.io/coder/coder:latest
curl -sf http://localhost:3000/api/v2/buildinfo   # wait until this succeeds

# 4. Bootstrap the first user (use an isolated CODER_CONFIG_DIR, or the web UI,
#    so this doesn't clobber your own default `coder` CLI session)
export CODER_CONFIG_DIR=/tmp/coderv2-manual-test
coder login http://localhost:3000 \
  --first-user-username admin --first-user-email admin@coder.com \
  --first-user-password "SomeSecurePassword123!" --first-user-trial=false
cat "$CODER_CONFIG_DIR/session"   # session token, paste into provider "token"
```

</details>

**Scenario A — creating a *new* template with `versions` omitted still fails clearly** (Coderd requires a version to exist at all):

```hcl
resource "coderd_template" "test" {
  name = "settings-only-template"
  # no `versions` block
}
```

<details>
<summary>Output: <code>terraform apply</code></summary>

```
Plan: 1 to add, 0 to change, 0 to destroy.
coderd_template.test: Creating...
╷
│ Error: Client Error
│
│   with coderd_template.test,
│   on main.tf line 14, in resource "coderd_template" "test":
│   14: resource "coderd_template" "test" {
│
│ At least one template version (with `active = true`) is required when
│ creating a new `coderd_template` resource, since Coder templates cannot
│ exist without a version.
│ To manage an existing template without Terraform-managed versions, use
│ `terraform import` instead.
╵
```

</details>

**Scenario B — create normally, then drop `versions` to `null`** — only the `versions` attribute should change, the template and its version stay put:

<details>
<summary>Output: create with one active version, then <code>versions = null</code></summary>

```
# terraform apply (with versions = [{ directory = "...", active = true }])
Plan: 1 to add, 0 to change, 0 to destroy.
coderd_template.test: Creating...
coderd_template.test: Creation complete after 2s [id=2490ed95-d47e-45cb-b033-a265479c046b]
Apply complete! Resources: 1 added, 0 changed, 0 destroyed.

# edit main.tf: versions = null, then terraform apply
  ~ resource "coderd_template" "test" {
        name                              = "settings-only-template"
      - versions                          = [
          - {
              - active         = true -> null
              - directory      = "/.../example-template" -> null
              - directory_hash = "f2a1566b..." -> null
              - id             = "50526510-8042-494a-b932-8c35639fb69c" -> null
              - name           = "gleaming_yang68" -> null
            },
        ] -> null
        # (17 unchanged attributes hidden)
    }
Plan: 0 to add, 1 to change, 0 to destroy.
Apply complete! Resources: 0 added, 1 changed, 0 destroyed.

# coder templates versions list settings-only-template
NAME             CREATED AT            CREATED BY  STATUS     ACTIVE
gleaming_yang68  2026-07-06T21:29:26Z  admin       Succeeded  Active
```

</details>

**Scenario C — the actual customer scenario: an external pipeline pushes and activates a new version while `versions` is `null`** — `terraform plan` must show *no drift*:

<details>
<summary>Contents of <code>example-template-2</code> (the "new" version the pipeline pushes)</summary>

`integration/template-test/example-template-2/main.tf`:
```hcl
variable "name" {
  type = string
}

resource "local_file" "a" {
  filename = "${path.module}/a.txt"
  content  = "hello ${var.name}"
}

output "a" {
  value = local_file.a.content
}
```

</details>

<details>
<summary>Output: <code>coder templates push --activate</code> outside Terraform, then <code>terraform plan</code></summary>

```
$ coder templates push settings-only-template -d ./example-template-2 --activate --yes
Updated version at Jul  6 14:30:25!

$ coder templates versions list settings-only-template
NAME                     CREATED AT            CREATED BY  STATUS     ACTIVE
frightening_robinson82  2026-07-06T21:30:24Z  admin       Succeeded  Active
gleaming_yang68          2026-07-06T21:29:26Z  admin       Succeeded

$ terraform plan
coderd_template.test: Refreshing state... [id=2490ed95-d47e-45cb-b033-a265479c046b]

No changes. Your infrastructure matches the configuration.

Terraform has compared your real infrastructure against your configuration
and found no differences, so no changes are needed.
```

</details>

**Scenario D — Terraform updates a setting while `versions` is `null`** — only the setting changes, the pipeline's active version must survive untouched:

<details>
<summary>Output: <code>terraform apply</code> with a new <code>description</code>, then re-check active version</summary>

```
$ terraform apply   # description = "now managed by an external pipeline" added
  ~ resource "coderd_template" "test" {
      ~ description                       = "" -> "now managed by an external pipeline"
        id                                = "2490ed95-d47e-45cb-b033-a265479c046b"
        name                              = "settings-only-template"
        # (16 unchanged attributes hidden)
    }
Plan: 0 to add, 1 to change, 0 to destroy.
Apply complete! Resources: 0 added, 1 changed, 0 destroyed.

$ coder templates versions list settings-only-template
NAME                     CREATED AT            CREATED BY  STATUS     ACTIVE
frightening_robinson82  2026-07-06T21:30:24Z  admin       Succeeded  Active
gleaming_yang68          2026-07-06T21:29:26Z  admin       Succeeded
```

`frightening_robinson82` (the pipeline's version) is still `Active` — Terraform never touched it.

</details>

### DIFF
--- a/internal/provider/template_resource.go
+++ b/internal/provider/template_resource.go
@@ -555,16 +555,9 @@ func (r *TemplateResource) Create(ctx context.Context, req resource.CreateReques
 		return
 	}
 
-	if len(data.Versions) == 0 {
-		resp.Diagnostics.AddError("Client Error",
-			"At least one template version (with `active = true`) is required when "+
-				"creating a new `coderd_template` resource, since Coder templates cannot "+
-				"exist without a version.\nTo manage an existing template without "+
-				"Terraform-managed versions, use `terraform import` instead.")
-		return
-	}
-	// The active-version requirement is enforced at plan-time by
-	// versionsPlanModifier.PlanModifyList, so it doesn't need to be repeated here.
+	// The "at least one version is required when creating" requirement is
+	// enforced at plan-time by versionsPlanModifier.PlanModifyList, so it
+	// doesn't need to be repeated here.
 
 	if data.OrganizationID.IsUnknown() {
 		data.OrganizationID = UUIDValue(r.data.DefaultOrganizationID)
@@ -1047,6 +1040,13 @@ func (d *versionsPlanModifier) PlanModifyList(ctx context.Context, req planmodif
 	// Returning here (without touching resp.PlanValue) leaves the planned
 	// value as-is, preserving null instead of coercing it into an empty list.
 	if req.ConfigValue.IsNull() || req.ConfigValue.IsUnknown() {
+		if req.ConfigValue.IsNull() && req.State.Raw.IsNull() {
+			resp.Diagnostics.AddError("Client Error",
+				"At least one template version (with `active = true`) is required when "+
+					"creating a new `coderd_template` resource, since Coder templates cannot "+
+					"exist without a version.\nTo manage an existing template without "+
+					"Terraform-managed versions, use `terraform import` instead.")
+		}
 		return
 	}
 
--- a/internal/provider/template_resource.go
+++ b/internal/provider/template_resource.go
@@ -17,6 +17,7 @@ import (
 	"github.com/coder/retry"
 	"github.com/coder/terraform-provider-coderd/internal/codersdkvalidator"
 	"github.com/google/uuid"
+	"github.com/hashicorp/terraform-plugin-framework-validators/listvalidator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/setvalidator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
 	"github.com/hashicorp/terraform-plugin-framework/attr"
@@ -469,6 +470,7 @@ func (r *TemplateResource) Schema(ctx context.Context, req resource.SchemaReques
 				MarkdownDescription: "The template versions to manage. If null, Terraform will not create, update, or read template versions, and will only manage the template's other settings. At least one version (with `active = true`) is required when creating a new template, since Coder templates cannot exist without a version.",
 				Optional:            true,
 				Validators: []validator.List{
+					listvalidator.SizeAtLeast(1),
 					NewVersionsValidator(),
 				},
 				NestedObject: schema.NestedAttributeObject{
@@ -561,16 +563,8 @@ func (r *TemplateResource) Create(ctx context.Context, req resource.CreateReques
 				"Terraform-managed versions, use `terraform import` instead.")
 		return
 	}
-	if hasActiveVersion, diag := hasOneActiveVersion(data.Versions); diag.HasError() {
-		resp.Diagnostics.Append(diag...)
-		return
-	} else if !hasActiveVersion {
-		resp.Diagnostics.AddError("Client Error",
-			"At least one template version must be active when creating a "+
-				"`coderd_template` resource.\n(Subsequent resource updates can be made "+
-				"without an active template in the list).")
-		return
-	}
+	// The active-version requirement is enforced at plan-time by
+	// versionsPlanModifier.PlanModifyList, so it doesn't need to be repeated here.
 
 	if data.OrganizationID.IsUnknown() {
 		data.OrganizationID = UUIDValue(r.data.DefaultOrganizationID)
@@ -1073,6 +1067,16 @@ func (d *versionsPlanModifier) PlanModifyList(ctx context.Context, req planmodif
 		return
 	}
 
+	// req.State.Raw.IsNull() is true only when there's no prior state at all,
+	// i.e. this is a genuine Create (unlike checking private state for nil,
+	// this stays false for a resource that was `terraform import`ed with
+	// `versions` omitted, since import populates state before this runs).
+	if req.State.Raw.IsNull() && !hasActiveVersion {
+		resp.Diagnostics.AddError("Client Error", "At least one template version must be active when creating a"+
+			" `coderd_template` resource.\n(Subsequent resource updates can be made without an active template in the list).")
+		return
+	}
+
 	for i := range planVersions {
 		hash, err := computeDirectoryHash(planVersions[i].Directory.ValueString())
 		if err != nil {
@@ -1089,11 +1093,6 @@ func (d *versionsPlanModifier) PlanModifyList(ctx context.Context, req planmodif
 		return
 	}
 	// If this is the first read, init the private state value.
-	// Note: this is also true for a resource that was `terraform import`ed
-	// with `versions` omitted from config (Create never ran for it), so we
-	// can't use "no prior private state" as a proxy for "this is being
-	// created" here — that check now lives in Create() itself, where it's
-	// unambiguous.
 	if lvBytes == nil {
 		lv = make(LastVersionsByHash)
 	} else {
--- a/internal/provider/template_resource.go
+++ b/internal/provider/template_resource.go
@@ -17,7 +17,6 @@ import (
 	"github.com/coder/retry"
 	"github.com/coder/terraform-provider-coderd/internal/codersdkvalidator"
 	"github.com/google/uuid"
-	"github.com/hashicorp/terraform-plugin-framework-validators/listvalidator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/setvalidator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
 	"github.com/hashicorp/terraform-plugin-framework/attr"
@@ -467,9 +466,9 @@ func (r *TemplateResource) Schema(ctx context.Context, req resource.SchemaReques
 				},
 			},
 			"versions": schema.ListNestedAttribute{
-				Required: true,
+				MarkdownDescription: "The template versions to manage. If null, Terraform will not create, update, or read template versions, and will only manage the template's other settings. At least one version (with `active = true`) is required when creating a new template, since Coder templates cannot exist without a version.",
+				Optional:            true,
 				Validators: []validator.List{
-					listvalidator.SizeAtLeast(1),
 					NewVersionsValidator(),
 				},
 				NestedObject: schema.NestedAttributeObject{
@@ -551,6 +550,25 @@ func (r *TemplateResource) Create(ctx context.Context, req resource.CreateReques
 	// Read Terraform plan data into the model
 	resp.Diagnostics.Append(req.Plan.Get(ctx, &data)...)
 	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if len(data.Versions) == 0 {
+		resp.Diagnostics.AddError("Client Error",
+			"At least one template version (with `active = true`) is required when "+
+				"creating a new `coderd_template` resource, since Coder templates cannot "+
+				"exist without a version.\nTo manage an existing template without "+
+				"Terraform-managed versions, use `terraform import` instead.")
+		return
+	}
+	if hasActiveVersion, diag := hasOneActiveVersion(data.Versions); diag.HasError() {
+		resp.Diagnostics.Append(diag...)
+		return
+	} else if !hasActiveVersion {
+		resp.Diagnostics.AddError("Client Error",
+			"At least one template version must be active when creating a "+
+				"`coderd_template` resource.\n(Subsequent resource updates can be made "+
+				"without an active template in the list).")
 		return
 	}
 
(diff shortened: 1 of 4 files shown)


internal/provider/template_resource.go
```diff
@@ -467,9 +466,9 @@ func (r *TemplateResource) Schema(ctx context.Context, req resource.SchemaReques
 				},
 			},
 			"versions": schema.ListNestedAttribute{
-				Required: true,
+				MarkdownDescription: "The template versions to manage. If null, Terraform will not create, update, or read template versions, and will only manage the template's other settings. At least one version (with `active = true`) is required when creating a new template, since Coder templates cannot exist without a version.",
+				Optional:            true,
 				Validators: []validator.List{
-					listvalidator.SizeAtLeast(1),
```
ethanndickson, line 472: This is compatible with `Optional`, so we should be able to keep it?

internal/provider/template_resource.go
```diff
@@ -1062,16 +1088,14 @@ func (d *versionsPlanModifier) PlanModifyList(ctx context.Context, req planmodif
 		resp.Diagnostics.Append(diag...)
 		return
 	}
-	// If this is the first read, init the private state value
+	// If this is the first read, init the private state value.
+	// Note: this is also true for a resource that was `terraform import`ed
+	// with `versions` omitted from config (Create never ran for it), so we
+	// can't use "no prior private state" as a proxy for "this is being
+	// created" here — that check now lives in Create() itself, where it's
+	// unambiguous.
```
ethanndickson: `req.State.Raw.IsNull()` is an unambiguous check as to whether the resource is being created, so we should be able to keep this check at plan-time, as part of the plan modifier. Having to move it out would is slightly worse UX. I have an example from the other day in `plan_modifiers.go`.

BobbyHo:
Thanks for the thorough review, Ethan. I really appreciate you taking the time, especially for catching the import implication in your second comment.

You’re right that req.State.Raw.IsNull() is the correct signal here. I hadn’t considered that an imported Terraform resource would hit the same “no prior private state” heuristic as a genuine Create, which could have caused the active-version check to silently misfire during import. Good catch.

I’ve updated the code based on both comments:

- Restored listvalidator.SizeAtLeast(1) alongside Optional: true on versions.
- Moved the active-version check back into the versions plan modifier, gated on req.State.Raw.IsNull() instead of the private-state heuristic, so it’s caught at plan time again and no longer misfires on import.

I also added acceptance test coverage for both cases, including a regression test that reproduces the import scenario and fails against the old heuristic. I re-ran the full manual test flow — plan/apply/import/destroy — against a real Coder deployment to confirm nothing broke.

Whenever you get a chance, I’d appreciate another look. 🙏

internal/provider/template_resource.go
```diff
@@ -554,6 +555,17 @@ func (r *TemplateResource) Create(ctx context.Context, req resource.CreateReques
...
+	if len(data.Versions) == 0 {
+		resp.Diagnostics.AddError("Client Error",
+			"At least one template version (with `active = true`) is required when "+
+				"creating a new `coderd_template` resource, since Coder templates cannot "+
+				"exist without a version.\nTo manage an existing template without "+
+				"Terraform-managed versions, use `terraform import` instead.")
+		return
+	}
+	// The active-version requirement is enforced at plan-time by
+	// versionsPlanModifier.PlanModifyList, so it doesn't need to be repeated here.
```
ethanndickson: Missed this the first time around, but we can catch this at plan time too, by checking the state value and the config value.

The state value being null tells us if this is a creation operation, the config value being null (alternatively; simply not unknown, they're the same thing in this case) tells us that the user has definitely omitted `versions`. You cannot rely on `len(data.Versions) == 0` at plan time because it might be unknown. (See agents.md, I've been documenting a bunch of these footguns)

internal/provider/template_resource.go
```diff
@@ -1030,6 +1042,14 @@ func (d *versionsPlanModifier) MarkdownDescription(context.Context) string {
 
 // PlanModifyObject implements planmodifier.List.
 func (d *versionsPlanModifier) PlanModifyList(ctx context.Context, req planmodifier.ListRequest, resp *planmodifier.ListResponse) {
+	// `versions` is optional: if it's null or unknown, Terraform is not
+	// managing template versions at all, so there's nothing to reconcile.
+	// Returning here (without touching resp.PlanValue) leaves the planned
+	// value as-is, preserving null instead of coercing it into an empty list.
+	if req.ConfigValue.IsNull() || req.ConfigValue.IsUnknown() {
+		return
+	}
```
ethanndickson, line 1051: re: my other comment

```suggestion
	if req.ConfigValue.IsNull() || req.ConfigValue.IsUnknown() {
		if req.ConfigValue.IsNull() && req.State.Raw.IsNull() {
			resp.Diagnostics.AddError("Client Error",
				"At least one template version (with `active = true`) is required when "+
					"creating a new `coderd_template` resource, since Coder templates cannot "+
					"exist without a version.\nTo manage an existing template without "+
					"Terraform-managed versions, use `terraform import` instead.")
		}
		return
	}
```

internal/provider/template_resource_test.go
```diff
@@ -679,6 +683,248 @@ func TestAccTemplateResource(t *testing.T) {
...
+			require.NoError(t, markActive(ctx, client, templateID, versionResp.ID))
+			pipelineVersionID = versionResp.ID
+		}
+
+		resource.Test(t, resource.TestCase{
+			PreCheck:                 func() { testAccPreCheck(t) },
+			IsUnitTest:               true,
+			ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
+			Steps: []resource.TestStep{
+				// Terraform creates the template and its first version, as usual.
```
ethanndickson:
```suggestion
```

internal/provider/template_resource_test.go
```diff
@@ -1295,8 +1541,12 @@ type testAccTemplateResourceConfig struct {
 	CORSBehavior                 *string
 	UseClassicParameterFlow      *bool
 
-	Versions []testAccTemplateVersionConfig
-	ACL      testAccTemplateACLConfig
+	// VersionsNull renders `versions = null` instead of a list, regardless of
+	// the contents of Versions. Used to test that Terraform can manage a
+	// template's settings without owning its version lifecycle.
+	VersionsNull bool
+	Versions     []testAccTemplateVersionConfig
```
ethanndickson: nit: *[]T would be better than a bool, also matches what I've done previously

internal/provider/template_resource_test.go
```diff
@@ -1374,6 +1624,48 @@ func (c testAccAutostopRequirementConfig) String(t *testing.T) string {
 	return buf.String()
 }
 
+// versionsString renders the `versions` attribute. When VersionsNull is set,
+// it renders `null` regardless of the contents of Versions, so tests can
+// exercise settings-only management of a template (see PLAT-288).
```
ethanndickson:
```suggestion

```

BobbyHo:
Thank you @ethanndickson for your suggestions again. I've updated the code and would you mind taking another look when you have a moment? Thanks!

internal/provider/template_resource_test.go
```diff
@@ -679,6 +683,245 @@ func TestAccTemplateResource(t *testing.T) {
...
+		resource.Test(t, resource.TestCase{
+			PreCheck:                 func() { testAccPreCheck(t) },
+			IsUnitTest:               true,
+			ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
+			Steps: []resource.TestStep{
+				{
+					Config: cfg.String(t),
+					// Terraform's CLI diagnostic renderer word-wraps long error
+					// text with real newlines, so `.` must match them too.
+					ExpectError: regexp.MustCompile("(?s)At least one template version.*is required when.*creating"),
```
ethanndickson, line 721: Can we mark this error as expected during *plan* by setting `PlanOnly`?

ethanndickson (approved):
Sick, thanks! One last comment

internal/provider/template_resource_test.go
BobbyHo: Thanks @ethanndickson, nice catch!. The latest commit includes PlanOnly: true in this test case.

PR #386, opened 2026-07-11
chore!: bump coder/coder go module to latest main

ethanndickson:
Bumps `github.com/coder/coder/v2` from the previously pinned main commit (`db7f4438b4ad`, 2026-07-01) to a newer `main` commit (`9af3d31036bf`, 2026-07-10).

Resulting module version: `v2.34.0-rc.0.0.20260710073612-9af3d31036bf`

## Breaking change

This SDK bump changes the `model_config` schema for `coderd_agents_model`, so this provider version requires a Coder deployment running `coder/coder` `main` (equivalently, Coder >= v2.36.0 once that is cut). It is **not** compatible with any currently released Coder.

- Reasoning effort moved out of per-provider options: `provider_options.anthropic.effort` and `provider_options.openai.reasoning_effort` are removed in favor of a top-level `reasoning_effort` object (`{ default, max }`). This restructure only exists on `coder/coder` `main`; the newest release branch (`release/2.35`) still uses the per-provider fields.

The example, generated docs, and integration fixtures are migrated to the new schema in this PR.

`go mod tidy` pulled in the corresponding transitive dependency updates. `go build ./...` passes.

Cost pricing (`model_config.cost`) is unchanged: it is nested under `cost` on every supported version and legacy top-level pricing keys are still folded in on read, so it is not affected by this bump.

## Release context

This is the second of **two releases**.

**Release A — shipped as [v0.0.21](https://github.com/coder/terraform-provider-coderd/releases/tag/v0.0.21).** The compatible bugfix release for released Coder (v2.35.x): clearer Bedrock `role_arn` error on older servers (#389, fixes #387), the empty-Bedrock-string null fix (#384), optional template `versions` (#383), and the computed Bedrock `external_id` (#382). It stays on the pre-bump SDK, so released-Coder users are unaffected by this PR.

**Release B — this PR.** The breaking SDK bump, compatible only with `coder/coder` `main` (Coder >= v2.36.0). To be cut and **marked breaking** once merged, with the minimum required Coder version called out in the release notes. Users on released Coder should stay on v0.0.21.

## Companion: #388

#388 adds a plan-time `model_config` validator that fails loudly on keys Coder would silently drop. Landing it alongside this bump turns a user's now-removed `provider_options.anthropic.effort` into a clear plan-time error (with the `effort` -> `reasoning_effort` migration hint) instead of a silent drop. It depends on this bump's SDK, so merge this first, then #388, and include both in Release B.

_Opened by Coder Agents on behalf of @ethanndickson._

PR #388, opened 2026-07-13
feat!: reject unrecognized model_config settings and bump coder/coder to latest main

ethanndickson:
Closes CODAGT-797

Adds a plan-time validator for `coderd_agents_model.model_config` that fails loudly when the config contains settings the provider would silently drop when decoding into its pinned Coder SDK.

## Why

Each provider release pins exactly one `codersdk`, and `model_config` is decoded into that SDK's `ChatModelCallConfig` before every create/update — so a release supports exactly one `model_config` schema. Standard `encoding/json` ignores object keys with no matching struct field, so anything outside that schema (a misspelled key, a field from a newer Coder, a field removed or renamed by an SDK bump) is silently dropped in the provider before the request is ever sent, with no plan diff, and the value in state compares semantically equal. A user can lose configured behavior with no signal.

## What

- New `agentsModelConfigNoDroppedKeysValidator`, wired into the `model_config` schema alongside the existing not-empty validator.
- It reports every dropped setting by its full dotted path (e.g. `provider_options.anthropic.bogus_setting`), so a name that is valid in one place cannot mask a dropped occurrence of the same name elsewhere.
- The error explains that the settings are not part of the schema this provider was built against, and tells the user to either upgrade the provider to a release built against their Coder version or fix/remove the field.

This is schema-change-agnostic: nothing in the implementation restates the SDK schema, so it cannot drift when `ChatModelCallConfig` evolves.

<details>
<summary>Design notes</summary>

- **The SDK is the sole oracle.** To decide whether a key is dropped, the validator removes it from the decoded document, re-canonicalizes through `ChatModelCallConfig`, and compares against the canonical output of the full document: a key whose removal leaves the output unchanged contributed nothing, so it was dropped. Nothing hardcodes recognized fields, legacy aliases, or folding rules.
- **Relocations and legacy aliases fall out for free.** A legacy top-level pricing key on its own (folded under `cost`) probes as *kept*, because removing it changes the output. The same key shadowed by a differing `cost.<key>` (whose nested value wins) probes as *dropped*, because removing the ignored legacy value changes nothing — so silent overrides are still caught.
- **Content-free values are skipped.** `null`, `[]`, and `{}` hit `omitempty` whether or not the key is recognized, so removing one never changes the output; the probe cannot judge them and does not try.
- `json.DisallowUnknownFields` does **not** work here: `ChatModelCallConfig` has a custom `UnmarshalJSON` that calls plain `json.Unmarshal` internally, discarding the decoder's strict setting for the whole subtree.
- This branch is independent of the coder/coder bump (PR #386): tests assert on synthetic unknown keys that drop on any SDK version.
</details>

## Compatibility model

The supported way to consume this (experimental) resource is to run the provider release that matches your Coder version; cross-version compatibility is delivered by the release stream (see #386's release plan), not by any single provider binary.

We considered making `model_config` an opaque passthrough blob (raw JSON both ways, bypassing the typed SDK client) so one provider release could span Coder versions, and rejected it as counter to that compatibility story. Passthrough would let this one attribute float across versions while everything else about the resource stays typed to the pinned SDK — a false sense of cross-version support — and it silently accepts typos and unsupported settings, which match no Coder version, get dropped by the server on write, and never come back: a perpetual diff with no error. This validator enforces the contract loudly instead: any setting outside the schema the provider was built against fails at plan time, with the remedy in the diagnostic (fix the field name, or upgrade the provider if the setting is valid on your Coder). Combining passthrough with this validation wouldn't help either — the error rejects exactly the newer-server fields passthrough exists to forward, and softening it to a warning just re-introduces quiet drift while adding the raw-HTTP machinery passthrough requires.

The validator checks against the pinned SDK, not the live server, so the reverse mismatch — a provider newer than your Coder — can still send fields the server drops; that too is solved by matching versions.

_Opened by Coder Agents on behalf of @ethanndickson._


ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

internal/provider/agents_model_config.go
```diff
@@ -232,3 +235,148 @@ func (v agentsModelConfigNotEmptyValidator) ValidateString(_ context.Context, re
...
+	return v.Description(ctx)
+}
+
+func (v agentsModelConfigNoDroppedKeysValidator) ValidateString(_ context.Context, req validator.StringRequest, resp *validator.StringResponse) {
+	if req.ConfigValue.IsNull() || req.ConfigValue.IsUnknown() {
+		return
+	}
+	// Invalid or non-object JSON is left for the custom type and the not-empty
+	// validator to report.
+	dropped, err := agentsModelConfigDroppedKeys(req.ConfigValue.ValueString())
```
johnstcn: Would it be simpler to just marshal to the SDK config type and back, and assert JSON equivalence? (leaving out the problems of ordering and whitespace)

johnstcn (commented):
I feel like this adds a lot of complexity for not much gain.

Do we plan eventually to promote the model config into a top-level Terraform type? Would it be simpler to just promote once it stabilizes?

ethanndickson:
> I feel like this adds a lot of complexity for not much gain.

Yeah I think that's a reasonable criticism. The reason I'm doing this is because Mike and his agent had no idea his coder/dogfood PR to adapt the thinking levels wasn't going to do anything until the codex review bot flagged the codersdk on this repo needed bumping, but I think this applies to customers too. People pin their Terraform providers, and without this they might upgrade coder, but forgot to update the provider, and be left wondering why their config didn't apply. The provider wouldn't tell them without this.

> Do we plan eventually to promote the model config into a top-level Terraform type? Would it be simpler to just promote once it stabilizes?

No plans to promote it unfortunately, when I looked at doing `model_config` as a top-level Terraform time it was hack after hack and hugely complex, thanks to things like fields typed with `map[string]any`. Also, I don't think it'll ever truly be stabilized? These AI providers seem to add and remove fields at the drop of a hat.

johnstcn:
> People pin their Terraform providers, and without this they might upgrade coder, but forgot to update the provider, and be left wondering why their config didn't apply. The provider wouldn't tell them without this.

Fair point. In that case, I wonder if it would make sense to move the validation into `codersdk`?

ethanndickson:
> > People pin their Terraform providers, and without this they might upgrade coder, but forgot to update the provider, and be left wondering why their config didn't apply. The provider wouldn't tell them without this.
> 
> Fair point. In that case, I wonder if it would make sense to move the validation into `codersdk`?

Ah good shout. Yeah we could just add an `UnmarshalStrict` to `ChatModelCallConfig` that sets `json.Decoder.DisallowUnknownFields`, and that would let us catch this with very little code.

internal/provider/agents_model_resource_test.go
```diff
@@ -391,6 +391,32 @@ func TestAgentsModelConfigNotEmptyValidator(t *testing.T) {
...
+
+	diags := validate(types.StringValue(`{"provider_options":{"anthropic":{"bogus_setting":"x"}}}`))
+	require.True(t, diags.HasError())
+	require.Contains(t, diags[0].Detail(), "bogus_setting")
+
+	require.False(t, validate(types.StringNull()).HasError())
+	require.False(t, validate(types.StringUnknown()).HasError())
+	// Invalid and non-object JSON are other validators' problem.
+	require.False(t, validate(types.StringValue(`{`)).HasError())
+	require.False(t, validate(types.StringValue(`[1,2]`)).HasError())
```
johnstcn: nit: could be refactored to a table test

internal/provider/agents_model_config.go
```diff
@@ -232,3 +233,46 @@ func (v agentsModelConfigNotEmptyValidator) ValidateString(_ context.Context, re
...
+	// The lenient decode above succeeded, so a strict failure can only be an
+	// unknown field.
+	var config codersdk.ChatModelCallConfig
+	if err := config.UnmarshalStrict([]byte(req.ConfigValue.ValueString())); err != nil {
+		resp.Diagnostics.AddAttributeError(
+			req.Path,
+			"Unrecognized model_config settings",
+			fmt.Sprintf(
+				"model_config contains a setting that is not part of the Coder chat model config schema this provider was built against, so it would be silently dropped before reaching Coder: %s. "+
+					"If the setting is valid for your Coder deployment, upgrade the provider to a version built against your Coder release; otherwise remove it or fix the field name. "+
```
johnstcn, line 272: non-blocking: As an end-user, how do you determine this? Something to think about in future, I suppose.

johnstcn (approved):

internal/provider/agents_model_config.go
ethanndickson: ~~ask an agent~~

Yeah, not sure... Bit of an experimental-API only paper cut... We haven't had to ask users to pay attention to the provider versions like this in the past.

ibetitsmike:
coder/dogfood#397 is now green and waiting on a provider release containing this SDK bump. Its current provider pin silently drops top-level `reasoning_effort`, which leaves one valid Codex finding unresolved. Once this PR lands and releases, I will bump dogfood, validate the rendered plan, and resume its Codex review loop.

> Mux posted this dependency note on behalf of Mike.

ibetitsmike:
The v0.0.22 GitHub release completed successfully, but Terraform Registry has not indexed v0.0.21 or v0.0.22. The Registry provider metadata is still `coder/coderd/0.0.20` with `last-modified: Wed, 01 Jul 2026`, and `terraform init` fails with `no available releases match the given constraints 0.0.22`. Release assets and signatures are present, and v0.0.20 through v0.0.22 use the same signing key.

Could a maintainer check or resync the Terraform Registry publication? coder/dogfood#397 is ready to consume v0.0.22 once the Registry serves it.

> Mux diagnosed and reported this release blocker on behalf of Mike.

ethanndickson:
> The v0.0.22 GitHub release completed successfully, but Terraform Registry has not indexed v0.0.21 or v0.0.22. The Registry provider metadata is still `coder/coderd/0.0.20` with `last-modified: Wed, 01 Jul 2026`, and `terraform init` fails with `no available releases match the given constraints 0.0.22`. Release assets and signatures are present, and v0.0.20 through v0.0.22 use the same signing key.
> 
> Could a maintainer check or resync the Terraform Registry publication? [coder/dogfood#397](https://github.com/coder/dogfood/pull/397) is ready to consume v0.0.22 once the Registry serves it.
> 
> > Mux diagnosed and reported this release blocker on behalf of Mike.

Yeah I'm resyncing now

PR #392, opened 2026-07-16
Make proxy`session_token` `Sensitive: true`

wyattfry:
Change schema to make proxy token sensitive so it does not show the value in Terraform logs.

Screenshot of the plan in TFE, it shows the token in the clear plaintext, I replaced real values with simulated ones:

image

With this change it should instead show "(sensitive)" or something like that.


ethanndickson (approved):
Good catch, thanks for the PR!

PR #400, opened 2026-07-27
fix(coderd_template): validate ACL roles

wyattfry:
This adds plan-time validation for coderd_template ACL roles so Terraform rejects unsupported values before apply. Valid roles are admin and use, matching the Coder SDK constants. It also updates generated template docs to describe the accepted ACL role values.

### DIFF
--- a/internal/provider/template_resource.go
+++ b/internal/provider/template_resource.go
@@ -258,7 +258,11 @@ var permissionAttribute = schema.SetNestedAttribute{
 				Required: true,
 			},
 			"role": schema.StringAttribute{
-				Required: true,
+				MarkdownDescription: "Role assigned to the user or group. Valid roles are `admin` and `use`.",
+				Required:            true,
+				Validators: []validator.String{
+					templateACLRoleValidator,
+				},
 			},
 		},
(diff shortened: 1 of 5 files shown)


internal/provider/template_resource.go
```diff
@@ -1179,6 +1183,11 @@ var weekValidator = setvalidator.ValueStringsAre(
 	stringvalidator.OneOf("monday", "tuesday", "wednesday", "thursday", "friday", "saturday", "sunday"),
 )
 
+var templateACLRoleValidator = stringvalidator.OneOf(
+	string(codersdk.TemplateRoleAdmin),
+	string(codersdk.TemplateRoleUse),
+)
+
```
ethanndickson, line 1190: Not blocking, but its unfortunate we have to define this list here, instead of importing it from codersdk.

ethanndickson (approved):

ethanndickson:
Thanks again!

internal/provider/template_resource.go
wyattfry: Agreed.

PR #408, opened 2026-08-11
feat(coderd_ai_provider): add bedrock protocol attribute

johnstcn:
## Problem

`coderd_ai_provider` has no `settings.bedrock.protocol` attribute, so HCL `protocol = "mantle"` is silently dropped. The API then receives a region-only provider with the zero protocol (`invoke-model`), which requires `model` and `small_fast_model`. This surfaced as a 400 when dogfood-prod tried to create a Mantle provider:

```
POST https://dev.coder.com/api/v2/ai/providers: 400
settings.model: model is required for the invoke-model protocol
settings.small_fast_model: small_fast_model is required for the invoke-model protocol
```

The Coder SDK already defines `AIProviderBedrockSettings.Protocol` (`invoke-model` default, `mantle`), `AIProviderBedrockProtocol{InvokeModel,Mantle}`, and `ResolvedProtocol()`. The provider just never exposed it.

Mantle (Coder v2.36.0+) is a passthrough protocol that forwards the native Messages request with SigV4 signing and passes the client's model through, so a single Mantle provider can serve many Anthropic models without `model`/`small_fast_model`.

## Fix

- Add `Protocol` to `AIProviderBedrockSettingsModel` and a `settings.bedrock.protocol` schema attribute: `Optional`, validated `OneOf("invoke-model", "mantle")`. Modeled on `role_arn` (the closest analogue: an optional capability with a server-version gate), so removal reverts to the default rather than being sticky.
- Thread it through every SDK-settings construction (`ValidateConfig`, `validateEffectiveUpdateState`, `sdkSettings`) and read it back in `stateFromProvider` (empty `""` maps to null, matching the omitted config).
- Add cross-field validation in `ValidateConfig`: `protocol = "mantle"` requires `region` (fail at plan time, since the Mantle base URL is not canonical and never derives a region). `Protocol` is part of the unknown-deferral guard.
- Add `checkBedrockProtocolDropped`, run after create/update: fails loudly if a configured protocol round-trips to nothing (older server predating v2.36.0 drops the unknown JSON key). Comparison normalizes an empty response to the legacy `invoke-model` default via `ResolvedProtocol()`.
- Update the example and regenerate docs.

## Tests

Unit (`internal/provider`):

- Schema validation rejects an invalid `protocol` and `protocol = "mantle"` without `region`; accepts the valid values.
- `createRequest` sends `Settings.Bedrock.Protocol == "mantle"`.
- `stateFromProvider` retains `protocol` (`mantle`, `invoke-model`, and empty→null).
- `checkBedrockProtocolDropped` errors when an old server strips it, passes when persisted or not configured, and treats an empty echo as the `invoke-model` default.
- Existing invoke-model/default tests remain green.

Integration (`integration/`):

- New `bedrock-protocol-test` spins up a Dockerized Coder and applies a Mantle provider (`protocol = "mantle"`, no `model`/`small_fast_model`), then asserts the protocol round-trips via the API and that no model fields are required. This mirrors the dogfood PR #434 scenario.
- The integration harness gained an optional `minVersion` gate (e.g. `v2.36.0`) that skips a test when the containerized Coder is older, since older servers drop the protocol field. The gate uses `BuildInfo().CanonicalVersion()` + `semver.Compare` (`golang.org/x/mod` promoted to a direct dependency, already in the module graph).

## Review

All four inline review threads are resolved. This change is generated by Coder Agents on behalf of the requester.

Follow-up dogfood work (bumping the provider pin and applying the seven Anthropic Mantle models) is tracked separately.


internal/provider/ai_provider_resource.go
```diff
@@ -371,6 +386,7 @@ func (r *AIProviderResource) ValidateConfig(ctx context.Context, req resource.Va
 			RoleARN:         bedrock.RoleARN.ValueString(),
 			AccessKey:       stringPtrOrNil(bedrock.AccessKeyWO),
 			AccessKeySecret: stringPtrOrNil(bedrock.AccessKeySecretWO),
+			Protocol:        bedrockProtocol(bedrock.Protocol),
```
ethanndickson, line 385: `protocol = "mantle"` with a missing `region` is a config-only invariant, but nothing rejects it until apply. `codersdk` returns a validation error for that pair, and `createReq.Validate()` runs inside `Create()`, so the operator sees a failed apply instead of a failed plan. AGENTS.md asks us to fail at plan time when the rule is decidable from config and prior state.

The gap matters more for Mantle than it looks. `bedrockCanonicalBaseURLRegex` matches only `bedrock-runtime.<region>.amazonaws.com`, so a Mantle base URL such as `https://bedrock-mantle.us-east-1.api.aws/anthropic` never derives a region. Every Mantle user must set `region` explicitly, and the current `region` description still promises derivation from `base_url`.

Two suggested changes:

1. Add the cross-field rule to `ValidateConfig`, next to the existing Bedrock checks:

```go
if bedrockProtocol(bedrock.Protocol) == codersdk.AIProviderBedrockProtocolMantle &&
    sdkSettings.Region == "" {
    resp.Diagnostics.AddAttributeError(
        path.Root("settings").AtName("bedrock").AtName("region"),
        "Missing Bedrock Region",
        "`protocol = \"mantle\"` signs requests with AWS SigV4 and requires `region`. The Mantle base URL is not a canonical Bedrock URL, so the region is not derived from `base_url`.",
    )
}
```

2. Note the exception in the `region` description, so the derivation promise stays accurate.

Generated by Coder Agents.

ethanndickson, line 385: The deferral guard above this block defers on unknown `Region`, `RoleARN`, `AccessKeyWO`, and `AccessKeySecretWO`, but not on `Protocol`:

```go
if !baseURLKnown || bedrock.Region.IsUnknown() || bedrock.RoleARN.IsUnknown() ||
    bedrock.AccessKeyWO.IsUnknown() || bedrock.AccessKeySecretWO.IsUnknown() {
    return
}
```

`bedrockProtocol` maps unknown to `""`, so an unknown `protocol` (from an input variable or a module output) is validated as invoke-model. This is harmless today, because `Protocol` does not affect `IsConfigured()`. It becomes a wrong verdict as soon as a protocol-dependent rule exists, for example the Mantle region rule. AGENTS.md records this class as "Required != known" (#368).

Please add `|| bedrock.Protocol.IsUnknown()` to the guard. A `TestStep.ConfigVariables` case with a required, no-default variable covers the validate walk, which is where the #305 family surfaced.

Generated by Coder Agents.

internal/provider/ai_provider_resource.go
```diff
@@ -250,6 +251,20 @@ func (r *AIProviderResource) Schema(ctx context.Context, req resource.SchemaRequ
...
+								Validators: []validator.String{
+									stringvalidator.OneOf(
+										string(codersdk.AIProviderBedrockProtocolInvokeModel),
+										string(codersdk.AIProviderBedrockProtocolMantle),
+									),
+								},
+								PlanModifiers: []planmodifier.String{
+									stringplanmodifier.UseStateForUnknown(),
+								},
+							},
```
ethanndickson, line 263: This shape copies `model` and `small_fast_model`, which is reasonable. It differs from `role_arn`, the closest analogue: an optional capability flag with a server-version gate. `role_arn` is `Optional` only, so deleting the line plans a null and clears the value.

With `Computed` plus `UseStateForUnknown`, deleting `protocol = "mantle"` produces **no diff**. The plan reuses the state value, and `sdkSettings` sends `mantle` again. The only way back to the default is to write `protocol = "invoke-model"` explicitly. Nothing in the schema description says so.

Please pick one:

- Keep the current shape and state the behaviour in the description, for example: "Once set, this value persists. Set `protocol = \"invoke-model\"` to return to the default; removing the attribute does not clear it."
- Drop `Computed` (and the plan modifier) to match `role_arn`, so removal clears the field. The server maps `""` to invoke-model, and `stateFromProvider` already maps `""` back to null, so removal round-trips cleanly.

Two related notes, both minor:

- `sdkSettings` reads `bedrock.Protocol` from the plan, while `region` prefers the config value with a plan fallback. The framework re-plans before apply, so this should resolve. It is still inconsistent with the neighbouring field.
- `checkBedrockProtocolDropped` compares for equality, while `checkBedrockRoleARNDropped` only checks for non-empty. Equality is correct while the server echoes the stored settings blob verbatim, which it does today.

Generated by Coder Agents.

ethanndickson (commented):
Reviewed the new `settings.bedrock.protocol` attribute against the sibling Bedrock attributes and against the footguns in AGENTS.md.

The attribute is correct on the two footguns that have burned this resource before: the `""` to null mapping (#384) and `UseStateForUnknown` on a derived value (#368). Three items need attention, one per inline comment:

1. The Mantle region rule fails at apply time, although it is decidable from config.
2. `Protocol` is missing from the unknown deferral guard in `ValidateConfig`.
3. `Optional + Computed + UseStateForUnknown` makes `protocol` sticky, unlike `role_arn`.

Two items outside the inline comments:

- `integration/bedrock-protocol-test/main.tf` has no trailing newline. `make fmt` runs `terraform fmt -recursive`, so `make fmt && git diff --exit-code` in the Definition of Done gate will fail.
- The PR promotes `golang.org/x/mod` from an indirect to a direct dependency for `semver.Compare`. AGENTS.md asks for approval before adding a dependency. The module is already in the graph, so the risk is low.

Generated by Coder Agents.

internal/provider/ai_provider_resource.go
johnstcn: Fixed. Added the cross-field rule to `ValidateConfig`: `protocol = "mantle"` with no `region` is now a plan-time `Missing Bedrock Region` error, since the Mantle base URL is not a canonical `bedrock-runtime` URL and never derives a region. Also amended the `region` description to note the Mantle exception to `base_url` derivation. Covered by the new `mantle protocol requires region` schema-validation case.

johnstcn: Fixed. Added `|| bedrock.Protocol.IsUnknown()` to the deferral guard so an unknown protocol defers validation instead of being treated as invoke-model. This is what makes the Mantle-region rule in `ValidateConfig` (which is protocol-dependent) correct for the unknown-variable case.

internal/provider/ai_provider_resource.go
```diff
@@ -675,6 +696,36 @@ func checkBedrockRoleARNDropped(config AIProviderResourceModel, provider codersd
...
+// checkBedrockRoleARNDropped).
+func checkBedrockProtocolDropped(config AIProviderResourceModel, provider codersdk.AIProvider, diags *diag.Diagnostics) {
+	b := config.bedrock()
+	if b == nil || b.Protocol.IsNull() || b.Protocol.IsUnknown() || b.Protocol.ValueString() == "" {
+		return
+	}
+	configured := b.Protocol.ValueString()
+	if provider.Settings.Bedrock != nil && string(provider.Settings.Bedrock.Protocol) == configured {
+		return
+	}
```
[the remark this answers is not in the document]
johnstcn: Fixed. `checkBedrockProtocolDropped` now compares via `ResolvedProtocol()`, so an empty response (the legacy invoke-model default) matches an explicitly configured `protocol = "invoke-model"` and no longer raises a false error. Added a `TestAIProviderCheckBedrockProtocolDropped` case: `invoke-model configured, server echoes empty (default)`.

internal/provider/ai_provider_resource.go
johnstcn: Chose option (b) for consistency with `role_arn`: `protocol` is now `Optional`-only with no plan modifier, so removing `protocol = "mantle"` clears it back to the default `invoke-model`. The server maps an empty value to `invoke-model` and `stateFromProvider` maps `""` back to null, so removal round-trips cleanly. The schema description now documents this behavior explicitly.

ethanndickson (approved):

PR #412, opened 2026-08-18
feat(coderd_chat_system_prompt): manage the deployment-wide chat system prompt

### LINKED ISSUE #411: Add a resource for the deployment-wide chat system prompt
## Problem

The deployment-wide chat system prompt (`Settings → Instructions` in the dashboard, `GET/PUT /api/experimental/chats/config/system-prompt`) can't be managed by this provider. It's one of the most frequently-changed pieces of Coder Agents config — it's effectively the org's institutional knowledge for agents — and right now the only way to manage it as code is to `curl` the API from CI.

That's exactly what we do in dogfood today for [cdrstable.dev](<http://cdrstable.dev>): `system-prompt.md` lives in git, the workflow `PUT`s it on apply and diff-warns on drift in PR plans. It works, but it's a workaround bolted onto a Terraform module that manages everything else natively:

* [https://github.com/coder/dogfood/blob/main/coderd-stable/README.md#system-prompt](<https://github.com/coder/dogfood/blob/main/coderd-stable/README.md#system-prompt>) (see "System prompt" — the README already says to fold this into the module "when a `coderd` resource for the system prompt ships")
* Workflow steps: [https://github.com/coder/dogfood/blob/main/.github/workflows/coderd-stable.yaml](<https://github.com/coder/dogfood/blob/main/.github/workflows/coderd-stable.yaml>)

## Everything needed already exists

* SDK client methods are in the pinned codersdk: `ExperimentalClient.GetChatSystemPrompt` / `UpdateChatSystemPrompt` ([codersdk/chats.go](<https://github.com/coder/coder/blob/main/codersdk/chats.go>))
* The provider already has singleton deployment-settings resources to copy the pattern from: `coderd_oauth2_provider_settings` (coder/terraform-provider-coderd#399) and `coderd_default_agents_model`
* It already uses the `ExperimentalClient` for `coderd_agents_model`, with the standard "experimental, changes expected" warning

## Proposed resource

```hcl
resource "coderd_chat_system_prompt" "this" {
  system_prompt                 = file("${path.module}/system-prompt.md")
  include_default_system_prompt = true # append to Coder's built-in prompt rather than replace
}
```

* Singleton (deployment-wide config, like `coderd_oauth2_provider_settings`): create/update both `PUT`, delete resets to empty prompt (or just forgets — open to opinions)
* Drift shows up in `plan` like any other resource, replacing the hand-rolled diff step
* Needs an `owner`-scoped token, same as the AI provider/model resources

## Related

* coder/coder#28090 — docs recipe for managing the system prompt as code; the "Follow-up" section there is this issue
* coder/dogfood#400 — the IaC PR that currently ships the workflow-PUT workaround
* Possibly overlaps with [PRODUCT-373](https://linear.app/codercom/issue/PRODUCT-373/expose-agent-settings-in-coderd-terraform-provider) (expose agent settings in the coderd Terraform provider) in Linear

> Generated by Coder Agents on behalf of @bpmct

bpmct:
Closes coder/terraform-provider-coderd#411

Adds `coderd_chat_system_prompt`, a singleton resource for the deployment-wide Coder Agents system prompt (`Settings → Instructions`). Replaces the CI `curl` workaround used in coder/dogfood#400.

```hcl
resource "coderd_chat_system_prompt" "this" {
  system_prompt                  = file("${path.module}/system-prompt.md")
  include_default_system_prompt = true
}
```

## Status

- [X] coder/coder#28283 merged (`codersdk.SanitizePromptText` exported)
- [X] SDK bump and model-cost migration split into coder/terraform-provider-coderd#413
- [X] Mirrored sanitizer deleted; this resource calls the SDK implementation directly
- [ ] Re-review @ethanndickson
- [ ] Merge + provider release, then coder/dogfood drops the workflow PUT/drift steps

## Design

* Coder sanitizes stored prompts, so `system_prompt` compares values by sanitized form (`codersdk.SanitizePromptText`). A trailing newline from `file(...)` never drifts; real edits still diff.
* Singleton like `coderd_oauth2_provider_settings`: one idempotent PUT, destroy resets deployment defaults, import adopts the live value, plan-time warnings before a first apply overwrites out-of-band values.
* Plan-time 128 KiB length validation.
* A 404 reports that the endpoint is unavailable and explains both likely causes: Coder older than v2.32.0 or insufficient site-wide permissions.

## Stack

Depends on coder/terraform-provider-coderd#414. The SDK export used by this resource is provided by the bottom stack layer, coder/terraform-provider-coderd#413.

1. coder/terraform-provider-coderd#413 — bump the Coder SDK and remove obsolete model cost settings
2. coder/terraform-provider-coderd#414 — expose `agents_allowed` on templates
3. coder/terraform-provider-coderd#412 — manage the deployment-wide chat system prompt

## Testing

* Real-Coder acceptance tests (`integration.StartCoder`): lifecycle no-drift with a deliberately messy prompt (CRLF, zero-width space, blank-line run, trailing newline), import convergence, destroy reset.
* Fake-coderd tests for request-level assertions (import issues no PUT, unavailable-endpoint diagnostic) plus unit tests for semantic equality, the validator, and the ModifyPlan advisories.
* Build, formatting, generated docs, lint, and provider unit tests pass locally on the restacked branch.

> Generated by Coder Agents on behalf of @bpmct


internal/provider/chat_system_prompt_sanitize.go
```diff
@@ -0,0 +1,123 @@
...
+//
+// The chat system prompt endpoint stores the sanitized form of
+// whatever is PUT to it, so the value read back rarely matches the
+// configured value byte-for-byte (a trailing newline from
+// `file("system-prompt.md")` is the everyday case). This local copy
+// exists so `system_prompt` can compare semantically: two values are
+// the same setting iff they sanitize to the same string. The logic is
+// deliberately a straight port; if the upstream sanitizer changes, the
+// worst case is a visible (loud) diff on the next plan rather than
+// silent drift.
```
ethanndickson: Instead of mirroring code from coder/coder, we should just export it as part of codersdk, and then we can just import it. I don't want to introduce drift. 

If we desperately want to get this resource in ASAP, I'm happy for this to be done as a followup.

Feel free to ping me for review on the coder/coder PR.

ethanndickson (changes requested):

internal/provider/chat_system_prompt_resource.go
```diff
@@ -0,0 +1,451 @@
...
+	if live.SystemPrompt == "" {
+		// Nothing configured out of band; nothing to lose.
+		return
+	}
+	if sanitizePromptText(live.SystemPrompt) == sanitizePromptText(data.SystemPrompt.ValueString()) {
+		// The planned prompt matches the live one.
+		return
+	}
+
+	resp.Diagnostics.AddAttributeWarning(
```
ethanndickson: We should also show the warning if `include_default_system_prompt` would be updated with the creation of the resource.

bpmct: Done in fa9771d: the create-time advisory now also fires when the first apply would change `include_default_system_prompt` away from the live value (message names both values). Covered by a direct `ModifyPlan` table test in the `oauth2_provider_settings` style, including the silent cases (greenfield, prior state exists, unknown planned value, failed lookup).

> Generated by Coder Agents on behalf of @bpmct

internal/provider/chat_system_prompt_sanitize.go
bpmct: Agreed on not mirroring — opened coder/coder#28283 exporting it as `codersdk.SanitizePromptText` (pure move + callsite updates, tests move with it) and requested you as reviewer.

For this PR I kept the local copy with a TODO pointing at that PR, since actually importing it requires bumping the pinned coder/coder here, and each dep bump is coupled to a release's `model_config` schema. So the swap lands as the follow-up you suggested: once 28283 merges and the next SDK bump happens, `chat_system_prompt_sanitize.go` gets deleted and the resource calls the codersdk function directly. The real-Coder acceptance test (applies a prompt with CRLF + zero-width space + blank-line run and requires an empty re-plan) is what keeps the copy honest in the meantime — if the server's sanitizer changes, that test breaks.

> Generated by Coder Agents on behalf of @bpmct

bpmct: https://github.com/coder/coder/pull/28283

bpmct: coder/coder#28283 merged, so this is done for real now: 21ac609 bumps the pinned coder/coder to `main@8a7e8d9d5b` and 70d90e3 deletes the mirror — semantic equality, the length validator, and the overwrite advisory all call `codersdk.SanitizePromptText` directly.

Heads-up on the bump: it crosses the cost-tracking removal (coder/coder@6b8f8204), so `ChatModelCallConfig` lost `Cost` and `UnmarshalStrict` now tolerates the removed pricing keys. The `agents_model` tests are adapted accordingly (details in the PR body's "SDK bump" section); full TF_ACC suite passes. If you'd rather own the bump with the release, I can split it out.

> Generated by Coder Agents on behalf of @bpmct

bpmct:
@ethanndickson ready for another review :)

ethanndickson (approved):
💯

PR #426, opened 2026-08-25
docs: group registry sidebar by subcategory

matifali:
## Problem

All `coderd_*` resources and data sources render on the Terraform Registry with an empty `subcategory`, so the sidebar lists every resource/data source flat with no grouping.

## Fix

Add per-resource doc templates under `templates/resources/<name>.md.tmpl` and `templates/data-sources/<name>.md.tmpl` (the [tfplugindocs conventional paths](https://github.com/hashicorp/terraform-plugin-docs#conventional-paths), matching the pattern used by hashicorp/aws, azurerm, google). Each is a copy of tfplugindocs' default template with a hardcoded `subcategory`, grouping pages into:

- **Agents** — `agents_mcp_server`, `agents_model`, `agents_system_prompt`, `agents_default_model`
- **AI Governance** — `ai_provider`
- **Deployment** — `license`, `oauth2_provider_settings`, `provisioner_key`, `workspace_proxy`
- **Identity & Access Management** — `organization`, `organization_group_sync`, `organization_sync_settings`, `group`, `user`
- **Templates** — `template`

The templates also drop the `HasImportIDConfig`/`HasImportIdentityConfig` branches from tfplugindocs' default template: no resource in this repo ships `import-by-string-id.tf`/`import-by-identity.tf` example files, so those branches were always-false dead code. Resource templates keep a single `HasImport` (`terraform import`) section; data-source templates have no import section (data sources are never imported).

Regenerated `docs/` with `make gen`; the diff against the base branch is exactly the `subcategory` line per file (19 files), nothing else changed.

## Stacking

Stacked on #423 (which stacks on #422). Covers the stack's org-scoped `coderd_agents_default_model` rename (subcategory: Agents); the base already includes the #425 `agents_system_prompt` rename.

## Note for future resources

With per-resource templates, a new resource/data source added without a matching `templates/` file falls back to tfplugindocs' embedded default (`subcategory: ""`) and renders uncategorized at the top of the sidebar. Copy a sibling template and set the right subcategory when adding one.

## Testing

- `make build`, `make fmt`, and `make gen` (idempotent, no drift) are all clean.
- `tfplugindocs validate --provider-name coderd` passes.
- `make lint` has 15 pre-existing `usetesting` findings, unchanged from the base (verified via `git stash`); unrelated to this change.


ethanndickson (approved):
I dedupe'd the tmpl files, lgtm

bpmct (approved):
Solid categorization

ethanndickson:
@codex review

PR #427, opened 2026-08-27
feat(coderd_agents_mcp_server)!: import by organization name and slug

ethanndickson:
Changes the `coderd_agents_mcp_server` import ID from `<organization-id>/<id>` to `<organization-name>/<slug>` (the organization component also accepts a UUID):

```shell
terraform import coderd_agents_mcp_server.example my-org/github-mcp
```

MCP server slugs are unique per organization, but the get-by-ID endpoint only accepts UUIDs, so import resolves the slug from the organization's server list. The resource hasn't shipped in a provider release, so the old UUID-composite format is dropped without a compatibility path.

Refs CODAGT-973.

Disclosure: Xum (AI agent) authored this PR on @ethanndickson's behalf.

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


ethanndickson:
Folded into #423 — this change now ships as part of that PR.

Issue #354, opened 2026-06-01
feat: Coderd Crossplane Provider


michvllni:
## Summary

It would be great to have a [Crossplane](https://www.crossplane.io/) provider that wraps this Terraform provider, so that Coder resources can be managed declaratively from within Kubernetes.

## Motivation

Today, managing Coder infrastructure (templates, template versions, users, groups, organizations, etc.) requires running Terraform as a separate step in the toolchain. For teams that already run their platform on Kubernetes and use GitOps (e.g. Flux/ArgoCD), this means context-switching between tools and reconciliation models.

A Crossplane provider would allow Coder and its data plane to be defined entirely as Kubernetes custom resources, continuously reconciled by Crossplane — no separate `terraform apply` pipeline required. This makes it possible to manage Coder the same way as the rest of the platform, all from a single control plane.

## Proposed approach

Crossplane has first-class tooling for generating providers directly from existing Terraform providers:

- **[Upjet](https://github.com/crossplane/upjet)** — Crossplane's code-generation framework that builds a native Crossplane provider from a Terraform provider's schema, mapping each Terraform resource to a Crossplane Managed Resource (CRD).

Because this repo is already a maintained Terraform provider with a well-defined schema, generating an Upjet-based Crossplane provider should be relatively straightforward and largely automated — the heavy lifting (CRD generation, controller scaffolding, Terraform invocation under the hood) is handled by Upjet. The main work tends to be:

1. Scaffolding the Upjet provider repo against this provider.
2. Configuring resource grouping / external-name conventions.
3. Handling provider configuration & credentials (Coder URL + token) via a `ProviderConfig`.
4. CI for regenerating when the underlying Terraform provider updates.

## Benefits

- Complete, declarative management of Coder **and** the data plane directly from Kubernetes.
- Native GitOps support without bolting Terraform into the pipeline.
- Single reconciliation model and toolchain for platform teams already on Crossplane.

PR #434, opened 2026-09-09
feat(fwprovider): export the provider implementation for embedding

michvllni:
Adds a public `fwprovider` package that hands out the provider implementation, so Go programs can embed the provider in-process instead of running it as a plugin binary.

## Motivation

In #354 I asked for a Crossplane provider for Coder. Crossplane's [Upjet](https://github.com/crossplane/upjet) generates one directly from a Terraform provider's schema, and it can drive that provider two ways:

- **Terraform CLI mode.** Upjet bundles the `terraform` binary and the provider binary into its controller image, then runs `terraform` against a workspace per managed resource, which in turn launches the provider plugin. It works against any provider and needs no cooperation from upstream.
- **In-process mode.** Upjet calls the provider's `provider.Provider` implementation directly through its plugin-framework client. No `terraform` binary, no provider binary, no subprocess per reconcile.

In-process mode needs an importable `provider.Provider`. `internal/provider` is unimportable outside this module, so today the only route is to fork this repo and add the export in the fork. That is exactly what Upbound does for the providers that keep theirs internal — [`terraform-provider-aws/xpprovider`](https://github.com/upbound/terraform-provider-aws) and [`terraform-provider-azurerm/xpprovider`](https://github.com/upbound/terraform-provider-azurerm), both wired in through a `replace` directive. Providers that expose theirs upstream need no fork at all: `terraform-provider-google` has [`google/fwprovider`](https://github.com/hashicorp/terraform-provider-google/tree/main/google/fwprovider), `terraform-provider-keycloak` has [`provider`](https://github.com/keycloak/terraform-provider-keycloak/tree/master/provider).

I would rather not maintain a fork of this repo that has to be re-synced on every release, and I doubt you want one circulating either.

## What changed

A single public package delegating to the existing constructor:

```go
func New(version string) func() tfprovider.Provider {
	return provider.New(version)
}
```

Returning the factory rather than an instance keeps the signature identical to `provider.New`, so it still drops straight into `providerserver.Serve`; embedders call it to get an instance.

Kept deliberately small:

- No new dependencies — `terraform-plugin-framework` is already required.
- No behaviour change and no second code path. It is a one-line delegation, so it cannot drift, and a signature change to `provider.New` breaks the build here too.
- `internal/` stays internal. This exports the provider surface only, not resource implementations, `codersdk` plumbing, or helpers.
- `main.go` is untouched.
- A test asserts the factory yields the provider `main.go` serves, with `TypeName`, version, resources and data sources all wired.

I named it `fwprovider` after the `terraform-provider-google` precedent rather than Upbound's Crossplane-specific `xpprovider`, on the grounds that this is an export you own rather than one living in a downstream fork. Happy to rename it, relocate it, or narrow the signature.

## Context

I have built an [Upjet-based Crossplane provider](https://github.com/axiansinfoma/crossplane-provider-coderd) covering all 15 resources, exposing both cluster-scoped and namespaced Crossplane v2 APIs, with a scheduled job that regenerates and opens a PR when a release lands here. It runs in Terraform CLI mode today, because that is the only mode available without forking this repo.

I took the liberty of starting the implementation rather than just filing the request, so there is something concrete to react to. If you would like the Crossplane provider itself to live under the Coder org, I am happy to donate it; equally happy to keep maintaining it out-of-tree if you would rather not adopt another repo. Either way this export stands on its own for anyone embedding the provider.

Refs #354

### DIFF
--- a/fwprovider/fwprovider_test.go
+++ b/fwprovider/fwprovider_test.go
@@ -1,0 +1,29 @@
+package fwprovider_test
+
+import (
+	"context"
+	"testing"
+
+	tfprovider "github.com/hashicorp/terraform-plugin-framework/provider"
+	"github.com/stretchr/testify/require"
+
+	"github.com/coder/terraform-provider-coderd/fwprovider"
+)
+
+// TestNew asserts that the exported factory yields the same provider main.go
+// serves, so embedders see the full set of resources and data sources.
+func TestNew(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.Background()
+	p := fwprovider.New("test")()
+	require.NotNil(t, p)
+
+	var resp tfprovider.MetadataResponse
+	p.Metadata(ctx, tfprovider.MetadataRequest{}, &resp)
+	require.Equal(t, "coderd", resp.TypeName)
+	require.Equal(t, "test", resp.Version)
+
+	require.NotEmpty(t, p.Resources(ctx))
+	require.NotEmpty(t, p.DataSources(ctx))
+}
--- a/fwprovider/fwprovider.go
+++ b/fwprovider/fwprovider.go
@@ -1,0 +1,22 @@
+// Package fwprovider exposes the coderd Terraform provider implementation to
+// other Go programs that embed it instead of running it as a plugin binary.
+//
+// The implementation lives in internal/provider, which the Go toolchain keeps
+// unimportable from outside this module. Embedders need a provider.Provider
+// value so they can read the schema and drive CRUD in-process; this package
+// hands them one without widening the rest of the internal API.
+package fwprovider
+
+import (
+	tfprovider "github.com/hashicorp/terraform-plugin-framework/provider"
+
+	"github.com/coder/terraform-provider-coderd/internal/provider"
+)
+
+// New returns a factory for the coderd provider implementation. The signature
+// matches what providerserver.Serve expects, so embedders can either serve it
+// as a plugin or call the factory directly to get an instance. version is what
+// the provider reports from its Metadata method.
+func New(version string) func() tfprovider.Provider {
+	return provider.New(version)
+}


fwprovider/fwprovider_test.go
ethanndickson, line 1: This doesn't look like it tests anything meaningful?

ethanndickson (commented):
`fwprovider` as a name is fine with me, and I don't have any objections with the export.

I'm not able to say if we'll have resources for any maintenance work this introduces, i.e. reviewing fixes for bugs only present downstream, but if this alone is useful I suppose it's fine?

fwprovider/fwprovider_test.go
michvllni: It tests that the provider is initialized via the factory by inspecting the metadata. I can add more thorough tests if you like

michvllni:
> `fwprovider` as a name is fine with me, and I don't have any objections with the export.
> 
> I'm not able to say if we'll have resources for any maintenance work this introduces, i.e. reviewing fixes for bugs only present downstream, but if this alone is useful I suppose it's fine?

I think that's fair. In the end this is an extension that allows extended usage. If there is something missing or issues, it would be the job of said extensions to handle it or open PRs to mitigate

fwprovider/fwprovider_test.go
ethanndickson: I don't think any tests are necessary here, my point was just that it doesnt seem possible for this test to fail without the provider just not working at all, and thus all the other tests failing.

michvllni: I agree, but as this just provides a public interface for the provider, I don't see how other tests could validate more. I mean I could create a test that validates the schema of the public interface with the one the provider internally provides, but I'm not sure if that will tell us more as, from my understanding, it will also always be true.
Having one baseline test that at least validates the creation should be okay in my opinion, but I can also remove that test alltogether if you prefer that

ethanndickson (approved):
Sounds good, thanks!

michvllni:
@ethanndickson could you merge the PR? I don't seem to have permission to do so. Thanks :)

ethanndickson:
Will make a new release with this

PR #438, opened 2026-09-14
chore: change `default_org_member_roles` to set

Emyrk:
Followup to https://github.com/coder/terraform-provider-coderd/pull/355#discussion_r4003853573.

Change `default_org_member_roles` on `coderd_organization` from a list to a set.

- Duplicate role names have no meaning, and ordering does not matter.
- Replaces the `minimum-implicit-member` experiment note in the attribute description with "Requires a Coder Deployment running v2.37.0 or later." The experiment was promoted to GA and removed in coder/coder#27472, first shipped in v2.37.0.
- Drops `minimum-implicit-member` from the acceptance test's `CoderExperiments(...)` list, since Coder v2.37.0+ no longer recognises it.
- Existing state migrates in place; Terraform converts list-of-string to set-of-string without a schema version bump.

---


internal/provider/organization_resource.go
```diff
@@ -151,11 +152,10 @@ This resource is only compatible with Coder version [2.16.0](https://github.com/
...
 
-			"default_org_member_roles": schema.ListAttribute{
+			"default_org_member_roles": schema.SetAttribute{
 				ElementType: types.StringType,
 				MarkdownDescription: "Built-in organization role names that are unioned into every member's effective roles. " +
-					"Changes propagate to members on their next request. Setting any value other than the deployment defaults " +
-					"requires the `minimum-implicit-member` experiment to be enabled on the Coder Deployment.",
+					"Changes propagate to members on their next request.",
 				Optional: true,
 				Computed: true,
```
ethanndickson, line 160: Ideally this resource should have a validator:

```
Validators: []validator.Set{
	setvalidator.ValueStringsAre(
		stringvalidator.OneOf("owner", "template-admin", "user-admin", "auditor"),
	),
},
```
and ideally we should import that list of strings from codersdk so it's not so brittle

ethanndickson: In saying that, we dont currently have an import for it on `coderd_user.roles`, which is where I copied that snippet from

Emyrk: > I actually don't think the session token from the provider block is available to us in the validator, so custom role validation is maybe impossible

We don't support custom roles here yet. So I can copy the list as it stands today. 
We don't add roles very often, but the list is only static to a given version of coderd.

ethanndickson:
@codex review

internal/provider/organization_resource.go
```diff
@@ -151,11 +152,10 @@ This resource is only compatible with Coder version [2.16.0](https://github.com/
 				},
 			},
 
-			"default_org_member_roles": schema.ListAttribute{
+			"default_org_member_roles": schema.SetAttribute{
 				ElementType: types.StringType,
 				MarkdownDescription: "Built-in organization role names that are unioned into every member's effective roles. " +
-					"Changes propagate to members on their next request. Setting any value other than the deployment defaults " +
-					"requires the `minimum-implicit-member` experiment to be enabled on the Coder Deployment.",
+					"Changes propagate to members on their next request.",
```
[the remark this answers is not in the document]
ethanndickson: I assume the experiment is just no longer required?

ethanndickson (approved):

PR #444, opened 2026-09-21
feat(internal/provider): accept OpenAI reasoning_mode in coderd_agents_model model_config

ibetitsmike:
Coder now supports OpenAI's Pro reasoning mode for chat models (coder/coder#29584): `provider_options.openai.reasoning_mode = "pro"` in the model config, for models from the GPT-5.6 Sol generation onward, including GPT-6 Astra.

## Problem

`coderd_agents_model.model_config` is validated against the provider's pinned `codersdk.ChatModelCallConfig` at plan time (#388), so with the current `github.com/coder/coder/v2 v2.37.1` pin any config that sets the new field fails before it is ever sent to Coder:

```
Error: Unrecognized model_config settings

model_config contains a setting that is not part of the Coder chat model config schema this provider was built against, so it would be silently dropped before reaching Coder: provider_options.openai.reasoning_mode. If the setting is valid for your Coder deployment, upgrade the provider to a version built against your Coder release; otherwise remove it or fix the field name.
```

## Fix

- Bump `github.com/coder/coder/v2` to `main` commit `a970c6c92e46` (`v2.34.0-rc.0.0.20260921160322-a970c6c92e46`), the same pseudo-version pattern as #388 and #413. Diffing the `ChatModelCallConfig` JSON tags between v2.37.1 and the new pin shows the `model_config` schema change is purely additive (`provider_options.openai.reasoning_mode` and `openai_config.reasoning_model`); no existing key is removed, so there is no migration and this is not a breaking release.
- Adapt test helpers to the renamed `codersdk.CreateUserRequest` (`OrganizationIDs` replaces `OrganizationID`, `DisableLogin` is gone). Provider code is unaffected.

No field-specific tests or examples are added for `reasoning_mode`; the existing validator and canonicalization tests cover the mechanism, matching how earlier dependency-bump fields were handled (review feedback, 0293fa2).

## Compatibility

Using `reasoning_mode` requires a Coder release that includes coder/coder#29584 (newer than 2.37); Coder validates the value server-side and OpenAI rejects it for models without Pro. `integration/agents-model-test` is intentionally left unchanged because the integration harness pulls `ghcr.io/coder/coder:latest`, which does not yet echo the field back.

> Xum opened this PR on behalf of @ibetitsmike.


internal/provider/agents_model_resource_test.go
```diff
@@ -411,6 +411,15 @@ func TestAgentsModelConfigCanonicalJSON(t *testing.T) {
...
 	})
 
+	t.Run("openai reasoning mode survives the round trip", func(t *testing.T) {
+		t.Parallel()
+		// Guards the pinned SDK: an older codersdk would silently drop the key here
+		// and the provider would plan a config Coder never receives.
+		got, err := agentsModelConfigCanonicalJSON(`{"provider_options":{"openai":{"reasoning_mode":"pro"}}}`)
+		require.NoError(t, err)
+		require.Equal(t, `{"provider_options":{"openai":{"reasoning_mode":"pro"}}}`, got)
+	})
```
ethanndickson:
```suggestion
```

Arguably redundant, we haven't done this for other fields we got via a dep bump

examples/resources/coderd_agents_model/resource.tf
```diff
@@ -33,3 +33,40 @@ resource "coderd_agents_model" "sonnet" {
...
+      default = "high"
+      max     = "xhigh"
+    }
+    provider_options = {
+      openai = {
+        reasoning_mode = "pro"
+      }
+    }
+  })
+}
```
ethanndickson:
```suggestion
```

docs/resources/agents_model.md
```diff
@@ -56,6 +56,43 @@ resource "coderd_agents_model" "sonnet" {
...
+      default = "high"
+      max     = "xhigh"
+    }
+    provider_options = {
+      openai = {
+        reasoning_mode = "pro"
+      }
+    }
+  })
+}
```
ethanndickson:
```suggestion
```

internal/provider/agents_model_resource_test.go
ibetitsmike: Removed in 0293fa2, along with the one-line `openai reasoning mode` case in the validator table, which had the same rationale.

> Xum replied on behalf of @ibetitsmike.

examples/resources/coderd_agents_model/resource.tf
ibetitsmike: Removed in 0293fa2; the example is back to main.

> Xum replied on behalf of @ibetitsmike.

docs/resources/agents_model.md
ibetitsmike: Removed in 0293fa2 by regenerating from the restored example (`make gen` is clean).

> Xum replied on behalf of @ibetitsmike.

ethanndickson (approved):

